PR #60, opened 2014-05-02
[WIP] - Adding ListenerDebugCommand

egulias:
There are 2 things in this PR
* A valid `composer.json` after passing the `composer validate` command
* The ListenersDebugCommand, which is a copy of the one used by the bundle since I'm not sure the command from the bundle would work adding it here https://github.com/egulias/DrupalAppConsole/blob/master/bin/console.php#L35

If this first implementation work, I can try requiring the bundle and using that command instead of copying the code.

@dmouse @jmolivas I would much appreciate if you can test this, I don't know Drupal :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
     "description": "Drupal 8 Console scaffolding module generator",
     "version": "@dev",
     "keywords": ["Drupal", "Console", "Development"],
-    "type": "library",
+    "type": "drupal-module",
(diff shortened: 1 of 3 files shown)


darol100:
Does still make sense to keep this PR alive ? Should we close it ? Does still apply ?

egulias:
Closing this until a way of having the containers dump is found. Will happily open it again by then.

PR #278, opened 2015-01-06
Enable add select, radios, checkboxes fields in form generator

enzolutions:
Enable add select, radios, checkboxes fields in form, with option to add options values via generator command


jmolivas:

PR #282, opened 2015-01-07
Translate to spanish some strings

rpayanm:


jmolivas:
Awesome @rpayanm thanks :)

PR #289, opened 2015-01-09
Rest resource command

enzolutions:


jmolivas:
Thanks for the PR @enzolutions ;)

PR #316, opened 2015-01-19
Update Readme

enzolutions:


jmolivas:
Thanks @enzolutions Not sure if we should we include this here or move to the gitbook repo what do you think @dmouse may be can include this with this PR by @danielnv18 on the documentation:
https://github.com/hechoendrupal/drupal-console-book/pull/22/files

PR #319, opened 2015-01-20
review config messages

esod:
I think "Load services form the container." should be "Load services from the container." The other changes I'm pretty sure about.


jmolivas:
Awesome thanks for the PR @esod

PR #321, opened 2015-01-21
Command to list modules on system

enzolutions:


jmolivas:
Merged, thanks for the PR and the new command

PR #324, opened 2015-01-22
Command module:install

enzolutions:


jmolivas:
Thanks for the PR and new command @enzolutions

PR #336, opened 2015-01-28
Mostly work on the block form.

esod:
For coding patterns I used these files:

core/modules/system/src/Plugin/Block/SystemMenuBlock.php
core/modules/book/src/Plugin/Block/BookNavigationBlock.php
modules/examples/block_example/src/Plugin/Block/ExampleConfigurableTextBlock.php
core/modules/aggregator/src/Plugin/Block/AggregatorFeedBlock.php


jmolivas:
Thanks for the PR @esod :)

PR #342, opened 2015-01-29
Removed unused variables

danielnv18:


danielnv18:
Removed the un-used use statement

PR #348, opened 2015-01-30
Generated files messages

jmolivas:


jmolivas:
This is how it looks like:
generate-entity-content

PR #361, opened 2015-02-03
generate:plugin:rest:resource systematically injects coding standard err...

martinfrances107:
My mistake I originally post this is an drupal.org issue 

https://www.drupal.org/node/2418827

Here is the text from the issue..

I have been using generate:rest:resource at bit recently which I really like

it is time to give a little back and fix some errors that are reported by

phpcs --standrard=Drupal .
Here is a sample of the coding standard bugs I have fixed.

Comments must end in a full stop
/* must become /**
comments have indentation errors.

Another obvious error is that the class definition need to be transformed from

class Hello extends foo
{
Anyway I hope this helps

class Hello extends foo {
but that is a system wide issue and needs to solved in another issue.


jmolivas:
Merged, thanks for the PR @martinfrances107 :)

PR #362, opened 2015-02-03
adds/fixes options for radios, select, and checkboxes in blocks

esod:
Checkboxes are not perfect (http://prntscr.com/60p8r5). I can take care of them in another PR. Before this PR, block form values were not printing in the view block at all.


esod:
These are the example files I based my changes on.

core/modules/book/src/Plugin/Block/BookNavigationBlock.php
core/modules/comment/src/CommentForm.php
core/modules/file/src/Plugin/Field/FieldWidget/FileWidget.php

Search for `'#type' => 'radios',`

PR #367, opened 2015-02-05
Remove the error: The "string" plugin does not exist

jjcarrion:
After generate:entity:content, when I added new fields to the entity using the UI I get this error:

"There was a problem creating field NameOfTheField: The "string" plugin does not exist."

I have seen that in the node entity the type in the form is string_textfield and not string.


jmolivas:
Merged, thanks for the PR @jjcarrion

jjcarrion:
thanks to you for the great work that you are doing! ;)

Issue #372, opened 2015-02-12
[rest:disable] Create command


omero:

jmolivas:
Fixed with #373

PR #373, opened 2015-02-12
#372 rest disable and refactor

omero:
@dmouse does a observation, and he suggest use `$this->getConfig Factory()` instead `\Drupal::config`


jmolivas:
Thanks for the PR @omero

PR #375, opened 2015-02-12
The list routes are now named collection routes.

bojanz:
Tracking core:

- Replaced every .list with .collection
- Added the collection link template to the entity annotations
- Used the new preferred way of setting redirects:
```
-    $form_state->setRedirect('entity.{{ entity_name }}.list');
+    $form_state->setRedirectUrl(${{ entity_name }}->urlInfo('collection'));
```
Note that I didn't do the same to the delete form's getCancelUrl(), I only fixed the route name there, since the idea is to remove that whole class in another PR (core now has generic delete forms).


jmolivas:
Thanks for the PR @bojanz

PR #399, opened 2015-02-17
#398 [config:override] New command

### LINKED ISSUE #398: [config:override] Create new command
We decide create a command that allow override a specific key in a config file for my drupal site:
`config:override system.site page.front node`

and the result will be something like that:

```
Configuration Original Value Updated Value
 page.front    node/1         node
```

omero:
This resolve #398


jmolivas:
Thanks for the PR @omero

PR #409, opened 2015-02-17
Adds permissions generator

esod:
This generator still needs to write its permissions into the routing.yml, but otherwise it does the job. 

Since this is new code and new functionality, I'm going to write a test for this generator next.


jmolivas:
Thanks for the PR @esod

PR #433, opened 2015-02-21
Remove dependencies and use Drupal Archive_tar class

enzolutions:


jmolivas:

PR #443, opened 2015-02-23
.install generator

esod:
This PR creates a .install file for modules that require a custom table. I need to write a test for this new functionality. Other than that several things have occurred to me while developing this generator.

Module generators need to work incrementally. Let's say I need to add a new column in the custom table, or a new field in the form, or a new permission. I don't want to have to re-generate that file from scratch again. I just want to add my new thing to the existing file and proceed. 

Anyway, this gets the .install generator on its way.

### DIFF
--- a/Test/Command/GeneratorInstallCommandTest.php
+++ b/Test/Command/GeneratorInstallCommandTest.php
@@ -1,14 +1,14 @@
 <?php
 /**
  * @file
- * Contains \Drupal\AppConsole\Test\Command\GeneratorPermissionCommandTest.
+ * Contains \Drupal\AppConsole\Test\Command\GeneratorInstallCommandTest.
  */
 
 namespace Drupal\AppConsole\Test\Command;
 
 use Symfony\Component\Console\Tester\CommandTester;
 
-class GeneratorPermissionCommandTest extends GenerateCommandTest
+class GeneratorInstallCommandTest extends GenerateCommandTest
 {
     /**
      * @dataProvider getInteractiveData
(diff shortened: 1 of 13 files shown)


esod:
The test is not working. Can someone tell me why I'm getting `Undefined offset: 5` on line 22 of GeneratorInstallCommandTest.php? Thanks.

jmolivas:
@esod may you merge it with master, in order to help you testing this

esod:
Yes I will, this evening, though.

jmolivas:
@esod can you try fixing conflicts, I really want to take a look at this PR and have it merged.

Crell:
I would actually question whether this generator is wise in the first place.  Most modules should not be creating their own DB schema, ever.  They should use KeyValue, State, Config, or Content Entities for data storage.  Custom SQL tables is an edge case in D8, even though it was common in the past.  Every custom SQL table makes supporting non-SQL cases more difficult.

Having a generator for creating a schema I fear will encourage bad habits.

esod:
Yup. I agree. This shouldn't be a part of the console. I wrote it when I was unaware of D8's ability to use other means for storage. Let's close the PR without merging it.

LowellMontgomery:
So many useful bits of information come from just "watching" DrupalConsole here on Github! Thanks for adding your 2¢, @Crell Definitely on the mark! :dart:

jmolivas:
Closing this since "Having a generator for creating a schema I fear will encourage bad habits"

jmolivas:
Thanks @esod for all of the work on this PR

Thanks @Crell for the clarification

PR #452, opened 2015-02-25
d8rules integration draft.

dasjo:


jmolivas:
Awesome @dasjo I will test and merge this. 

Do you think it will be any chance to have this integration at rules module as Webprofiler module ?
https://github.com/lussoluca/webprofiler/tree/8.x-1.x/src/Command

PR #456, opened 2015-02-26
Use the official drupal repository.

joshuataylor:
This uses the official drupal repository. Need discussion for this.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
     },
     "require-dev": {
         "drupal/coder": "dev-8.x-2.x",
-        "hechoendrupal/drupal": "8.0.0-beta6",
+        "drupal/drupal": "8.0.0-beta6",
         "phpunit/phpunit": "4.4.*"
     },
     "minimum-stability": "dev",


jmolivas:
What do you guys think ? @dmouse @enzolutions @omero 
Should we use repo from drupal or github ?

enzolutions:
Beta 7 is not supported yet right, I don't remember the original reason we didn't use at the begging was something related with travis @dmouse ?

joshuataylor:
Actually... drupal/drupal is supported by packagist, and submitted by Dries.

https://packagist.org/packages/drupal/drupal

No reason not to use it?

dmouse:
Yes, drupal/drupal is available on Packagist, drupal/drupal only have one dependency, this dependency is drupal/core. 

If you use the package drupal/drupal, you cloned all history of drupal only for read the composer.json

If you use the drupal/core you cloned the core directory with the vendor directory, this is unnecessary. 

For this reasons we have created the repo only with core directory without vendors.

jmolivas:
I think we should update top Beta7, I am testing the commands and I am almost fix the issues I found => https://github.com/hechoendrupal/DrupalAppConsole/issues/455 and this change will be need it.

Related to use `drupal/core`, `drupal/drupal` or keep `hechoendrupal/drupal` I have not mush to say since @dmouse is how have been taking care of travis config.

jmolivas:
Probably the downside keep using `hechoendrupal/drupal`is that we should maintain and tagging releases on the repo https://github.com/hechoendrupal/drupal on every `drupal/drupal` release

webflo:
There is already a semi-official subtree split https://packagist.org/packages/drupal/core maintained by @tstoeckler. It is with vendor but this should not matter for your use-case.

jmolivas:
@dmouse set it to `"drupal/core": "8.0.0-beta7"` I think this PR should me closed now what do you guys think about ? @dmouse @enzolutions @joshuataylor @webflo

webflo:
Yup, its good.

PR #466, opened 2015-02-28
Wrong message for missing module name when uninstalling a module

czettnersandor:


jmolivas:
Thanks for the PR @czettnersandor

PR #476, opened 2015-03-01
[config:export] Export site configuration results in a PHP Fatal error

hugronaphor:
This piece of code:
```
$this->addSuccessMessage(
  $output,
  sprintf($this->trans('commands.config.export.messages.directory'), $config_export_file)
);
```
is throwing an error:
```
PHP Fatal error:  Call to undefined function 
Symfony\Component\Console\Formatter\_drupal_get_last_caller() in 
/var/www/DrupalAppConsole/vendor/symfony/console/Symfony/Component/Console/
Formatter/OutputFormatter.php on line 37
```

It seems like $output var is obsolete in addSuccessMessage() as a string is need to be passed.

### DIFF
--- a/config/translations/console.en.yml
+++ b/config/translations/console.en.yml
@@ -61,7 +61,7 @@ commands:
       arguments:
         directory: Define export directory to save configuration output.
       messages:
-        directory: Your configuration export file was stores in %s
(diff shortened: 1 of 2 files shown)


jmolivas:
Thanks for the PR @hugronaphor

Issue #478, opened 2015-03-02
[console] Use annotations in Commands


dmouse:
The Drupal commands interact with more stuffs that the commands in Symfony, in Drupal, one command can be have many dependencies with other modules or themes or any other scope in your installation.

This approach is to use Annotations to provide metadata on each command to specify dependencies or check another variables in Drupal or any other stuff that interacts with the console.

For example:
```php
/**
 * @DrupalCommand (
 *      dependencies = {
 *          "simpletest"
 *      },
 * )
 */
class TestDebugCommand extends ContainerAwareCommand { ... }
```

enzolutions:
+1

jmolivas:
Are we still trying to add this @dmouse ? any advantage vs the current command discovery method ?

darol100:
@dmouse , Can you explain why you added the tag "core-issue"? I'm just curios.

darol100:
@dmouse , @jmolivas Shall we close this issue ? This is done.

PR #479, opened 2015-03-02
Annotations in Drupal Commands #478

dmouse:
Follow the discussion in #478


jmolivas:
@dmouse Do you have any plans to keep working on this?, do you think adding this will add a value vs current file command discovery method.

greg-1-anderson:
I was previously unaware of this effort. See #2081.

Issue #381, opened 2015-02-13
[helper] Improve validation for Class name


langelhc:
I was creating a command and when I get promted:
     Enter the Command Class name [DefaultCommand]: qwer%w
I typed "%" and also typed an space for this value and I didn't get an error message.

This is the RegEx located in Validators.php:
     `REGEX_CLASS_NAME = '/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+$/' `

I was reviewing this with https://www.regex101.com/
*\x7f-\xff a single character in the range between the following two characters
\x7f the character  with position 0x7f (127 decimal or 177 octal) in the character set
\xff the character ÿ with position 0xff (255 decimal or 377 octal) in the character set (case sensitive)*

hugronaphor:
The RegEx by itself works as intended as recommended on http://php.net/manual/en/language.oop5.basic.php .
Validation was missing.
Created a PR:
https://github.com/hechoendrupal/DrupalAppConsole/pull/488

hugronaphor:
Out of topic question:
What coding standards do we follow?

jmolivas:
Since this project is not a Drupal module, the application code must follow:
* http://www.php-fig.org/psr/psr-1/
* http://www.php-fig.org/psr/psr-2/

Bu the generated result is a drupal module, that code must follow:
* https://www.drupal.org/coding-standards

hugronaphor:
Jesus @jmolivas 
Make sense.

jmolivas:
This is an issue related to what you just mentioned @hugronaphor 
* Clean up output code - https://github.com/hechoendrupal/DrupalAppConsole/issues/447

* Implement PSR2 coding standard for code-base - https://github.com/hechoendrupal/DrupalAppConsole/pull/400

PR #488, opened 2015-03-03
#381: Validate input class-name for generate:Command and PluginBlock.

hugronaphor:
Validate the Class name for Command and PluginBlock generation.
+ Describe correct Class name rules as on php.net


jmolivas:
I am working no have this PR working updating the test @hugronaphor

Issue #493, opened 2015-03-05
[generate:module] DefaultController routing.yml  


jmolivas:
Copied from d.o. https://www.drupal.org/node/2446347

Posted by keboca on March 4, 2015 at 8:39pm

First, congrats your Drupal integration with Symfony2 Console is a great idea made true. Now I have been checking and I noticed some particular cases related with routing built by console when it generated a default controller for a module.

When we execute: "drupal generate:module", some kind of assistant asks a questions related with our new module, one of those is about the Default Controller:
"Do you want to generate a default Controller [no]? "
If we accepted it, console builds a new Router.yml and Controller files.

The routing is defining "path" property without a leading forward slash, I aware the path will work as expected, but I suggest add the "/" prefix to keep the Drupal standard, basically all core module are doing that.
https://www.drupal.org/node/2092643

Another case is related with "defaults" property, it sets a "_controller" where defines the new "hello" method of our "DefaultController", but when you are trying to access to default path (custom_page/hello/{name}), Drupal will throws the following message:
"Not Acceptable"

It's because when Drupal\Core\EventSubscriber\ViewSubscriber class is executing "onView" method after checked the request type cannot handles the response.

It issue was kicking my tail, but after read excellent blog: [UNDERSTANDING DRUPAL 8, PART 3: ROUTING](https://cipix.nl/understanding-drupal-8-part-3-routing), I modified the "_controller" to "_content" then it starts to work properly, so I high recommend modify it when Drupal Console is building the routering file.

Anyway if you keep the "_controller" key, you must to modify the "hello()" method to return a Response object.

kenneth-bolivar-castro:
@jmolivas, I have tested drupal console using Drupal 8 beta 7, and it's handling the "_controller" key implemented to "defaults" property, then console could keep current behavior.

Also I checked the way how console is building the routing.yml file, and I found that Drupal\AppConsole\Generator\ModuleGenerator class executes "generate" method. Then I am able to prepend a leading forward slash at line 69.

I noticed, I haven't enough permission over project, then I couldn't create a PR, so I leave you the patch URL, please take a look and let me know if it works for you?

https://drive.google.com/file/d/0B2YsMp2BH2ShWTNwM0ZoaUNIWnc/view?usp=sharing

jmolivas:
I added a related issue #494 to fix the same on `generate:controller`

PR #494, opened 2015-03-06
#493 DefaultController routing.yml prepend a leading forward slash

kenneth-bolivar-castro:
DefaultController routing.yml prepend a leading forward slash to path property


jmolivas:
Awesome @kenneth-bolivar-castro, since you are working on this issue may you care about doing the same fix on the `ControllerGenerator` 
Probably asking if `$route` value does not start with `/` then prepend it 
https://github.com/hechoendrupal/DrupalAppConsole/blob/master/src/Generator/ControllerGenerator.php#L20

jmolivas:
Thanks for the PR @kenneth-bolivar-castro 
I added a related issue #495 to fix the same on `generate:controller`

Issue #495, opened 2015-03-07
[generate:controller] Prepend a leading forward slash on route path


jmolivas:
Add leading forward slash `/` on route definition

kenneth-bolivar-castro:
@jmolivas I added a new pull request, please take a look,

PR #496, opened 2015-03-08
#495 Prepend leading forward slash on route path

kenneth-bolivar-castro:
Check if $route param has leading forward slash.


jmolivas:
Thanks for the PR @kenneth-bolivar-castro

kenneth-bolivar-castro:
no problem, thanks to all of you to let me be part of this great project,

PR #502, opened 2015-03-10
Add a default form handler.

benjy:
The default entity template should add a default form handler.


jmolivas:
Thanks for the PR @benjy

Issue #506, opened 2015-03-12
[generate:permission] Missing params and some feedback after the generation


omero:
Looking the `generate:permissions`, and comparing with other modules, i see some usefull params.
```yaml
access content:
   title: 'Access Content'
   description: 'Allow access to my content.'
   restrict access: true
```
It will be nice to give some feedback to user, something like; 
```
You can use this permission in the routing file modules/custom/demo/demo.routing.yml example:
node.content:
   requirements:
     _permission: 'access content'
```
I do not think it's necessary ask if you want generate permissions that's the reason to call this command, but we can ask if you want to generate other permission.

Also i note that the command overwrite the permission.yml file, we can make an append, it will be responsability if the user remove the permissions, if they don't use it.

jmolivas:
Closing this since PR was merged. Feel free to re-open @omero if something is missing.

PR #509, opened 2015-03-13
Enhancement permission generator

omero:


omero:
PR for #506  
if i'm not wrong @esod wrote this command, do you have any suggestions, something we can do better or something isn't necessary ?

jmolivas:
Looks good to me any comments ?

esod:
Nice. The changes improve the flow of the script and it's great to see the script making additions to existing permissions.yml files. 

I think `restrict access` should be either `true`, `false`, or `none`. `none` is something I'm working on with the install generator if I can ever get the test to pass...

On line 53 of PermissionTrait.php
- $title = $this->getStringUtils()->camelCaseToUcFirst($title);
+ $title = $this->getStringUtils()->anyCaseToUcFirst($title);

camelCaseToUcFirst() doesn't exist unless it's in another branch. I wrote `anyCaseToUcFirst()` in StringUtils.php for the permission_title. I thought `title:` needed to be identical to the permission except for the UC first, so I used application code to generate it. But I see in `book.permissions.yml` that the permission and the title don't need to be identical:

I can't be sure, exactly, since I only knew to cut and paste the changes into my local console.dev, but I'd check the twig templates to make sure the indents are correct.

omero:
Thanks for the comments @esod if you are working in the option for the `none` value, it would be nice adding when you contribute this feature, like the integration when @jmolivas wrote the `-- lerning` param. 

`camelCaseToUcFirst()` exist in the master branch https://github.com/hechoendrupal/DrupalAppConsole/blob/master/src/Utils/StringUtils.php#L86 , I not see the `anyCaseToUcFirst()` but i will rename the function in this PR, I agree that the function can apply for `any`, and not exclusive for `camel case`

omero:
Thanks for the comments @esod, i did the changes that you suggest ... `camelCaseToUcFirst()` exist in the master branch https://github.com/hechoendrupal/DrupalAppConsole/blob/master/src/Utils/StringUtils.php#L86 , I not see the `anyCaseToUcFirst()` but i rename the function in this PR, I agree that the function can apply for any, and not exclusive for camel case

jmolivas:
Thanks for the PR @omero and the provided help @esod

Issue #513, opened 2015-03-18
[Violation] Text files should end with a newline character - in templates/…/permission-routing.yml.twig, line 9


danielnv18:
**[in templates/module/permission-routing.yml.twig, line 9](https://github.com/hechoendrupal/DrupalAppConsole/blob/ffe14b2b0312a6ae21ade494e626526d3bfd0383/templates/module/permission-routing.yml.twig#L9)**

> This file ends with no newline character. It won't render properly on a terminal, and it's considered a bad practice. Add a simple line feed as the last character to fix it.

```jinja
  {% for permission in permissions %}
  {{ module_name }}.route:
      requirements:
        _permission: {{ permission.permission }}:
  {% endfor %}
{% endif %}
```

PR #514, opened 2015-03-18
[Insight] Text files should end with a newline character - in templates/...

danielnv18:
#513


jmolivas:
Thanks for the PR @danielnv18

PR #523, opened 2015-03-23
Updated translations and file compared and sync with en version

kendallsv:


jmolivas:
Thanks for the PR @kendallsv

PR #537, opened 2015-03-28
French traduction

kgaut:


jmolivas:
Thanks for the PR @kgaut

Issue #533, opened 2015-03-27
generate:entity:content file annotation is incorrect - module account?


martinfrances107:
I love this project .... but using it I have found a bug.

I was using it to create a content entity type 

console generate:entity:content

with the following input 

module name: flag
entity class name: flagging
entity name flagging

everything works fine it creates a bundle of new files 

looking at the FlaggingAccessController.php file 

I see a tiny flaw, to report

/**
 * @file
 * Contains Drupal\account\FlaggingAccessController.
 */

the file annotation is wrong the project account is incorrect. it should be user defined ..in this case flag

This is just a bug report for the moment ... this weekend is busy, but next weekend I will have a look at tracking it down further...

martinfrances107:
Structure of project was so cleanly laid out - the bug was easy to find and it  is a simple one line fix. :)

jmolivas:
Thanks @martinfrances107 
You can send a PR from here:
https://github.com/hechoendrupal/DrupalAppConsole/compare/master...martinfrances107:master

vacho:
I found that this issue could be easily solved .. hehehe ...
https://github.com/hechoendrupal/DrupalAppConsole/pull/544

jmolivas:
This was already solved with https://github.com/hechoendrupal/DrupalAppConsole/pull/542

jmolivas:
Thanks for sending the PR @vacho

PR #544, opened 2015-03-30
entity_content: solved static module name 'account' in file class commen...

vacho:
this commit solve the issye #533


jmolivas:
This was previously fixed with https://github.com/hechoendrupal/DrupalAppConsole/pull/542 thanks for the PR @vacho

PR #564, opened 2015-04-05
Update README.md

Chi-teck:
`//` is not valid comment syntax in shell scripts


jmolivas:
Thanks for the PR @Chi-teck

PR #577, opened 2015-04-14
Change type from drupal-module to project.

greg-1-anderson:
The Console is no longer a Drupal module, so its type should be project.


jmolivas:
Thanks for the PR @greg-1-anderson

Issue #582, opened 2015-04-15
[chain] Multiple modules in one chain file.


Jaesin:
Is it possible to generate two different modules using a chain manifest file: 

This manifest will try to create the first module twice.
```yaml
# test.chain.yml
commands:
  chain:
    - generate:module
    - generate:module
  generate:
    module:
      options:
        module: Test Module
        machine-name: test_module
        module-path: modules/sandbox/
        description: Test Stuff.
        core: 8.x
        package: test
        controller: false
        test: false
    module:
      options:
        module: Test Module UI
        machine-name: test_module_ui
        module-path: modules/sandbox/
        description: User Interface for the Test Module.
        core: 8.x
        package: test
        controller: false
        dependencies: test_module
        test: false
```

Results in : `Unable to generate the bundle as the target directory ".../modules/sandbox/test_module" is not empty`

Jaesin:
After reading a bit of code and it looks like you are only able to have one of each command because of the way the [ChainCommand](https://github.com/hechoendrupal/DrupalAppConsole/blob/master/src/Command/ChainCommand.php#L80) accesses the options and arguments.

As a result, you can only have one say `generate:service` or `generate:plugin:block` command the way it is.

jmolivas:
Yes you are right, any ideas how to improve it ?

I will take a look at this tomorrow on my flight back.

Jaesin:
I would change the file format to something like: 

```yaml
# test.chain.yml
commands:
  - command: generate:module
    options:
        module: Test Module
        machine-name: test_module
        module-path: modules/sandbox/
        description: Test Stuff.
        core: 8.x
        package: test
        controller: false
        test: false
    arguments:
        - arg1
        - arg2
  - command: generate:module
      options:
        module: Test Module UI
        machine-name: test_module_ui
        module-path: modules/sandbox/
        description: User Interface for the Test Module.
        core: 8.x
        package: test
        controller: false
        dependencies: test_module
        test: false
    arguments:
        - arg1
        - arg2
```

Then you can simplify the execute loop: 
```php
/**
 * @file
 * Contains \Drupal\AppConsole\Command\ChainCommand.
 */

$chainData = new Config($file);
$commands = $chainData->get('commands');

foreach ($commands as $command) {
    $moduleInputs = [];
    $arguments = !empty($command['arguments']) ? $command['arguments'] : [];
    $options = !empty($command['options']) ? $command['options'] : [];

[...]

PR #584, opened 2015-04-15
#582: [chain] Allow multiple versions of the same command. fixes ...

jmolivas:
...#582.


jmolivas:
Thanks for the PR @Jaesin

PR #604, opened 2015-04-17
[generate:entity:content] Prevent "Link templates accepts paths" exception with i18n

kgaut:
Prevent "Link templates accepts paths, which have to start with a leading slash" exception when enabling content translation module

When you enable content translation module, and create a translatable entity with console, an exception is thrown by EntityType because there is no trailing slash at the begining of the path.


jmolivas:
Thanks for the PR @kgaut

PR #611, opened 2015-04-22
Fix parameters order in generate() for rest plugin generation.

jeqq:
At the moment a rest plugin is generated with incorrect id and label, the order of these parameters is wrong.


jmolivas:
Thanks for the PR @jeqq

PR #612, opened 2015-04-24
Add composer.json to ModuleGenerator

sethfischer:
Create composer.json in the module root.

### DIFF
--- a/Test/Command/GeneratorModuleCommandTest.php
+++ b/Test/Command/GeneratorModuleCommandTest.php
@@ -15,14 +15,14 @@ class GeneratorModuleCommandTest extends GenerateCommandTest
      */
     public function testInteractive($options, $expected, $input)
     {
-        list($module, $machine_name, $dir, $description, $core, $package, $controller, $dependencies, $test) = $expected;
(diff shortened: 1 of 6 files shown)


jmolivas:
Thats for the PR @sethfischer, I like the idea, how about adding an argument and interactive question on the module generator class to generate only if user pass `--composer` inline or answer yes to something like `Do you want to add a composer.json file to your module ?`

sethfischer:
Great idea @jmolivas. I'll update the PR.

jmolivas:
@sethfischer: Feel free to ask any question commenting in this thread or using the project chat https://gitter.im/hechoendrupal/DrupalConsole

sethfischer:
@jmolivas: I've updated this PR so that composer.json is only generated if the option `--composer` is specified, or the user responds yes to the interactive question "Do you want to add a composer.json
file to your module?". I welcome your comments on my implementation.

jmolivas:
Looks fine to me

jmolivas:
Thanks for the PR @sethfischer

PR #630, opened 2015-05-27
Update rest resource template

enzolutions:


jmolivas:
I will fix the extra spaces and merge this @enzolutions

PR #631, opened 2015-05-27
Enable provide multiple migration ids inline and interactive

enzolutions:


jmolivas:

PR #651, opened 2015-06-07
Require autoload.php instead core/vendor/autoload.php

webflo:
Should fix some issues with drupal-composer from drupal-composer/drupal-project/issues/28


jmolivas:
Thanks for the PR @webflo, I tested and it works great.

PR #657, opened 2015-06-08
PSR-1 and PSR-2 Coding Standards with .php_cs

ysramirez:
Following the PSR-1 and PSR-2 ;)


jmolivas:
Thanks for the PR @ysramirez 

I rather like to have this PR only with the `.php_cs` file and instructions how to test and then we decide how to apply the fixes.

PR #685, opened 2015-06-13
Support dev drupal as well

tarjei:
By changing the requirement from 2.6.* to >=2.6.* we get support for the dev version of Drupal that is installed with the https://github.com/drupal-composer/drupal-project project.


jmolivas:
We may create a `develop` branch to keep the work in sync with drupal-dev and test latest drupal changes.

webflo:
@jmolivas guzzle 6 broken the composer import. Its not possible to install drupal core and console in a single composer file because drupal core requires guzzle 5.

jmolivas:
@webflo I sent and merged a PR https://github.com/hechoendrupal/DrupalConsole/pull/754 with this fix

jmolivas:
I think this should be closed, any thoughts ?

webflo:
@jmolivas Yes it fine but. But as i wrote in https://www.drupal.org/node/2538484#comment-10149668 we should tag a console release now with guzzle 5. This release will work with older betas (pre beta13) and bump guzzle to guzzle 6 afterwards.

darol100:
@jmolivas , Do you still want to make sure that we support dev ? I think we should only support stable version because I think is going to add more complexity of the project.

jmolivas:
Closing this for now, since we are going to focus on support stable Drupal 8 releases.

PR #700, opened 2015-06-17
Set proper package type in composer.json

webflo:
The naming convention for composer package types is ``drupal-module``, ``drupal-theme``, ``drupal-profile``etc.


jmolivas:
Thanks for the PR @webflo

Issue #663, opened 2015-06-10
[generate:entity:*] Generated entity links should point to paths, not routes


Crell:
A recent core change requires that the links array in the Entity annotation be paths, NOT route names.  The current code is still generating an annotation with route names, which causes mysterious failures in some other code, such as devel. :smile: 

Very soon the routes themselves will not need to be generated as they will be automatic based on the annotation.  For right now, the path just needs to be double-listed.

jmolivas:
Thanks for the notice @Crell, do you have a link to see what we need to change?

Crell:
Here's the relevant change record: https://www.drupal.org/node/2382937

heilop:
I want to work in this issue.

jmolivas:
Awesome thanks @heilop please take your time to read this documentation 
http://hechoendrupal.gitbooks.io/drupal-console/content/contributing/new-features.html

And then try to complete the setup, and fork/clone and remote upstream and fee free to ping me at 
https://gitter.im/hechoendrupal/DrupalConsole

Sutharsan:
It looks like this is already fixed in e03a5240.

jmolivas:
This was fixed with PR https://github.com/hechoendrupal/DrupalConsole/pull/702

PR #702, opened 2015-06-17
#663 Generated entity links should point to paths, not routes

heilop:
I changed point to paths.


webflo:
@heilop Prefix ``/node`` is wrong. Please use the same url patterns as in https://github.com/hechoendrupal/DrupalConsole/blob/master/templates/module/routing-entity-content.yml.twig

heilop:
Thanks @webflo, I will change that.

jmolivas:
Thanks for the PR @heilop

heilop:
Thanks for help me @jmolivas 
Thanks @webflo for the code review 
yay!!

PR #706, opened 2015-06-25
Remove fieldable from ContentEntity annotation

webflo:
Related change record: https://www.drupal.org/node/2346455


jmolivas:
Thanks for the PR @webflo nice catch.

PR #709, opened 2015-06-25
Add weight to entity delete task

webflo:
... otherwise the Delete tab is shown before Edit.


jmolivas:
Thanks for the PR @webflo make sense :)

PR #710, opened 2015-06-25
Fix access operation in AccessControlHandler

webflo:
The operation is called "update" instead of "edit". Related docs from "EntityAccessControlHandler::checkAccess"

```
   * @param string $operation
   *   The entity operation. Usually one of 'view', 'update' or 'delete'.
```


jmolivas:
Thanks for the PR @webflo

Issue #717, opened 2015-06-29
[generate:rules*] Configuration objects exsist when reinstall module


omero:
### Problem/Motivation
After generata an rules action plugin with `generate:plugin:rulesaction` creates the `system.action.default_action.yml` configuration, when uninstall the module, the configuration object doesn't removes, and when i reinstall the module this happens:

![screen shot 2015-06-29 at 12 24 12 am](https://cloud.githubusercontent.com/assets/1909779/8401518/3058f47e-1df5-11e5-86ff-83845de01d49.png)

### How to reproduce

* Run `generate:plugin:rulesaction` in a new module.
* Uninstall the module
* Reinstall the module

### Solution
Include the custom module as dependency of the config object, when the module uninstall the configuration object will be erased too.

jmolivas:
Thanks @omero

PR #718, opened 2015-06-29
adding the module name has dependency of the configuration object template

omero:
solves #717


jmolivas:
Thanks for the PR @omero

PR #724, opened 2015-07-01
Remove deprecated default_value key from entity_value

artfulrobot:
The [correct way](http://drupal.stackexchange.com/questions/162287/how-to-create-a-default-value-for-an-entity-base-field-in-drupal-8) to set a default value on an entity is to use `setDefaultValue()` and **not** to pass a `default_value` key in the base table defs.

See related [patch to examples module](https://www.drupal.org/node/2511182).

Nb. I think it is useful to keep this code (which sets a zero-length string as the default) in there as an example of how to provide a default.


jmolivas:
Thanks for the PR @artfulrobot

PR #726, opened 2015-07-01
Convert t() to $this->t()

stevector:
This PR updates templates generated classes that still used t().

I'm chatting with Tim Plunkett in IRC right now about https://github.com/hechoendrupal/DrupalConsole/blob/master/templates/module/module.twig

Which generates

```
/**
 * Implements hook_help().
 */
function dcdtest_help($route_name, RouteMatchInterface $route_match) {
  switch ($route_name) {
    // Main module help for the dcdtest module.
    case 'help.page.dcdtest':
      $output = '';
      $output .= '<h3>' . t('About') . '</h3>';
      $output .= '<p>' . t('My Awesome Module') . '</p>';
      return $output;

    default:
  }
}
```

### DIFF
--- a/templates/module/src/Entity/Controller/listcontroller-entity-content.php.twig
+++ b/templates/module/src/Entity/Controller/listcontroller-entity-content.php.twig
@@ -26,8 +26,8 @@ class {{ entity_class }}ListController extends EntityListBuilder {% endblock %}
    * {@inheritdoc}
    */
   public function buildHeader() {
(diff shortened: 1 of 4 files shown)


stevector:
Tim says, and I agree, that t() in procedural code is fine. It does nothing but wrap  \Drupal::translation()->translate()

jmolivas:
Thanks for the PR @stevector

PR #745, opened 2015-07-10
Improve module generator options to controller

enzolutions:


jmolivas:
Discarding this one and move this new features to Controller generator

Issue #722, opened 2015-06-30
[generate:entity:config] Default menu path doesn't match up


Crell:
When generating a config entity, the path of the admin forms is placed under /admin/config/system/{entity}.  The menu LINK, however, is parented to /admin/structure.

While that technically doesn't break anything, it's highly confusing. The generator should standardize on one or the other.  I nominally think it makes more sense under structure than config/system, but as long as it's consistent I can deal.

Sutharsan:
PR at https://github.com/hechoendrupal/DrupalConsole/pull/756

PR #756, opened 2015-07-20
Set config entity path to admin/system/[entity name]

Sutharsan:
This makes the config entity path consistent. See  https://github.com/hechoendrupal/DrupalConsole/issues/722


Sutharsan:
I don't see how this Travis error is related to the included change.

jmolivas:
@Sutharsan: I restarted few times already but keep failing, I will test the code locally and add the results as comments.

jmolivas:
I did run it again and now it's green.

jmolivas:
@Sutharsan: Thanks for the PR

PR #759, opened 2015-07-21
Add human readable name

Sutharsan:
This PR adds a human readable name (--label) to content entity and config entity generators.
Adds a human readable formatter `StringUtils::camelCaseToHuman`.
Applies the HR label format to plugins.

Affected commands:
```
generate:entity:config
generate:entity:content
generate:plugin:block
generate:plugin:field
generate:plugin:fieldformatter
generate:plugin:fieldtype
generate:plugin:fieldwidget
generate:plugin:imageeffect
generate:plugin:rest:resource
generate:plugin:rulesaction
generate:plugin:type:annotation
```

### DIFF
--- a/config/translations/console.en.yml
+++ b/config/translations/console.en.yml
@@ -179,10 +179,12 @@ commands:
         module: common.options.module
         entity-class: The entity class name
         entity-name: The name of the entity
+        label: The entity label
       questions:
         module: common.questions.module
         entity-class: Enter the entity class name
         entity-name: Enter the entity name
+        label: Enter the entity label
     form:
       description: Generate a new "%s"
       help: The <info>"%s"</info> command helps you generate a new "%s"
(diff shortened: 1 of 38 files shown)


jmolivas:
Thanks for the PR @Sutharsan

pfrenssen:
This now looks a bit weird. If I create a "Book" entity type the permission titles now reads:

* View book
* Edit book
* Delete book
* Adminster book

Because this is singular this is confusing. Which book am I allowed to edit? What if I want to give a role the permission to edit all books?

It would be better if we follow the pattern used for all permissions in core: display the plural form. Since we do not know how our entity is spelled in plural we can simply add the word "entities":

* View book entities
* Edit book entities
* Delete book entities
* Adminster book entities

In the future it might be interesting to ask the user for the plural form when generating entities, we might be able to use it elsewhere too. But for now this would do the trick.

When we are touching this it would also be good to fix the spelling errors (e.g. "adminster") and remove the descriptions, since they are not adding any additional information and have bad grammar.

I'll open a new PR for this.

PR #761, opened 2015-07-22
Allow guzzle 5 and 6

webflo:


jmolivas:
Project already using `~5.0` https://github.com/hechoendrupal/DrupalConsole/blob/master/composer.json#L40

I build a new `phar`and it works, I just need to tag a new release.

dawehner:
It is IMHO bullshit to try to support two versions ... see https://www.drupal.org/node/2538484

Let's switch to 6.0

webflo:
I close the PR, because it makes no sense to discuss this in two different issue queues.

PR #763, opened 2015-07-24
New generic option to generate inline call of commands

enzolutions:


jmolivas:

PR #764, opened 2015-07-28
[console] Compact php files in the phar archive

Chi-teck:
This will decrease size of _console.phar_ by 40% approximately.

### DIFF
--- a/box.json
+++ b/box.json
@@ -18,6 +18,9 @@
             "in": "vendor"
         }
     ],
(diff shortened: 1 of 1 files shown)


jmolivas:
This sounds really awesome, I give a try and add comments if needed.

jmolivas:
@Chi-teck: thanks for the PR, the `phar` size went from 4.7 to 2.7 MB

PR #766, opened 2015-07-31
[console] Add database dump command

davidDuymelinck:
As promised i've added the beta 12 database dump command to drupal console.

### DIFF
--- a/config/translations/console.en.yml
+++ b/config/translations/console.en.yml
@@ -125,6 +125,14 @@ commands:
       messages:
         service_id: Service ID
         class_name: Class Name
+  database:
+    dump:
+      description: Dump the current database to a generation script
+      options:
+        file: File to write the generated script to
+      messages:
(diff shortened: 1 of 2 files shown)


darol100:
@davidDuymelinck, this seem like a very nice feature to have. For some reason it fails the automate test.

Can you try to re-roll your PR ? This will be something that will be very nice to have.

jmolivas:
@darol100 & @davidDuymelinck we added `database:dump` command, at this point only supports `mysql` because we used `mysqldump` command instead of iterating Database tables.

Command class link https://github.com/hechoendrupal/DrupalConsole/blob/master/src/Command/Database/DumpCommand.php

darol100:
@jmolivas I'm going to open another stories to handle other types of databases.

darol100:
Support for other database see more at #1559

Issue #743, opened 2015-07-08
[generate:entity:config] Separate class name and human name in wizard


Crell:
When generating a config entity (or probably content entity, but I've not tried) that logically is multiple words, the generated code uses the class name as if it were the human-readable name.  That's not the case, however.  Eg, if I'm creating a ModerationState config entity, I end up with strings and labels and page titles like "Edit ModerationState", "Add new ModerationState", etc.  So the very first thing I have to do is go through and change all of those labels to add a space... which is exactly what a generator is for. :smile: 

Instead, the wizard should ask for a class name and, separately, a human-friendly name (defaulting to the class name), and use the latter in all user-facing strings.

jmolivas:
Like the idea.

Sutharsan:
Working on this...

Sutharsan:
Done. See #759

jmolivas:
Fixed

PR #768, opened 2015-08-02
[generate:entity:config] Improve spelling and grammar where the new entity labels are used

pfrenssen:
Since #743 went in the permissions for generated content entities now look a bit weird. If I create a "Book" entity type the permission titles now read:

* View book
* Edit book
* Delete book
* Adminster book

Because this is singular this is confusing. Which book am I allowed to edit? What if I want to give a role the permission to edit all books?

It would be better if we follow the pattern used for all permissions in core: display the plural form. Since we do not know how our entity is spelled in plural we can simply add the word "entities":

* View book entities
* Edit book entities
* Delete book entities
* Adminster book entities

In the future it might be interesting to ask the user for the plural form when generating entities, we might be able to use it elsewhere too. But for now this would do the trick.

When we are touching this it would also be good to fix the spelling errors (e.g. "adminster") and remove the descriptions, since they are not adding any additional information and have bad grammar.

### DIFF
--- a/templates/module/entity-content-page.php.twig
+++ b/templates/module/entity-content-page.php.twig
@@ -4,7 +4,7 @@
 {{ entity_name }}.page.inc{% endblock %}
 {% block extra_info %}
  *
- * {{ entity_class }} page callback file for the {{ label }}.
+ * Page callback for {{ label }} entities.
 {% endblock %}
 
 {% block use_class %}
--- a/templates/module/links.menu-entity-content.yml.twig
+++ b/templates/module/links.menu-entity-content.yml.twig
@@ -2,11 +2,11 @@
 entity.{{ entity_name }}.collection:
   title: '{{ label }} list'
   route_name: entity.{{ entity_name }}.collection
-  description: 'List {{ label }}'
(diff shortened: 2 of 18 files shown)


pfrenssen:
This also relates to pull request #759.

pfrenssen:
While combing through the code I found some more instances where the human readable labels are now causing weird grammatical structures, some examples:

* Book page callback file for the Book.
* Defines the Book.
* Access controller for the Book.
* The language code of Book.
* ...

There were some instances where the label was preceded with 'a', this doesn't work well if we have entities that start with a vowel (e.g. 'Author'):

* Provides an interface defining a Author.
* This template is used when viewing a Author page.

I also updated the machine names of the permissions to also end in 'entities', this is consistent with how core does it (plural). Something like {{administer book entities}} is clearer than {{administer book}}. This will also help to avoid collisions with other permissions that happen to start with 'add/edit/delete/administer' but are not dealing with entities.

jmolivas:
@pfrenssen:  Thanks for the PR, reads much better

pfrenssen:
Thanks!

PR #779, opened 2015-08-06
Remove closing bracket from FieldFormatter template

webflo:
The closing bracket for the class is part of ``base/class.php.twig``.


jmolivas:
@webflo: Thanks for the PR

PR #780, opened 2015-08-06
Generated FieldFormatters, FieldTypes and FieldWidgets are wrong directory

webflo:


jmolivas:
@webflo: Thanks for the PR

Issue #781, opened 2015-08-07
Chasing HEAD: generate:plugin:field no longer functions


heddn:
- [x] There are some obvious syntax errors in the result of `drupal generate:plugin:field` i.e. semi-colons in the wrong place, missing parenthesis.
- [x] The directories for the generated classes does not include Field for all classes created
  ex. src/Plugin/FieldFormatter (correct is src/Plugin/Field/FieldFormatter
-  [x] Lastly, the field formatter references an used import: `use Drupal\Core\Field\FieldDefinitionInterface;`

heddn:
The above PR doesn't fix all the issues. I don't know how to solve the second bullet item. But it gets us closer.

jmolivas:
@heddn: Thanks this PR https://github.com/hechoendrupal/DrupalConsole/pull/780 fix the point number 2

heddn:
The second bullet from above is still not fixed in HEAD. I just built a new PHAR and tested. I'll see if I can figure out why.

The directories for the generated classes does not include Field for all classes created ex. src/Plugin/FieldFormatter (correct is src/Plugin/Field/FieldFormatter

jmolivas:
According to the merged PR looks like should be fixed https://github.com/hechoendrupal/DrupalConsole/pull/780/files
But looking at the code not fixed yet, feel free to send a PR to have it fix.

jmolivas:
@heddn: You can see what is required to have it fix by looking at the code of the mentioned PR
https://github.com/hechoendrupal/DrupalConsole/pull/780/files

enzolutions:
@heddn could yo confirm if generate:plugin:field is now working for you?

heddn:
It was a few weeks ago.

PR #782, opened 2015-08-07
Issue #781: Improves usability of Field generation

heddn:


templates/module/src/Plugin/Field/FieldFormatter/fieldformatter.php.twig
```diff
@@ -43,6 +43,7 @@ class {{ class_name }} extends FormatterBase {% endblock %}
    * {@inheritdoc}
    */
   public function settingsForm(array $form, FormStateInterface $form_state) {
+    $elements = array();
```
jmolivas: May we use show syntax array 
```
$elements = [];
```

heddn:
This changes things around a little to make the boilerplate that is generated hit the 80% target. Most of the time I'm creating a field, I want a text field. If this isn't the target desired, I can scale things back and leave only the shell.

heddn:
The last commit fixes some issues with the annoations. It was never printing the field_types in the annotations for a field generator. Some of the strings were copy/pasted for formatter using the language from a widget. And #780 doesn't seem to have merged for some reason. So, re-adding that functionality.

heddn:
OK, this PR is ready for merge now. I'm going to work on something else next.

jmolivas:
@heddn: Thanks I will give it a review and test and the merge. ping you if any questions arise.

heddn:
As a bonus to testing this, I uncovered https://www.drupal.org/node/2552799.

jmolivas:
@heddn: Thanks for the PR

PR #783, opened 2015-08-09
769 clear failed login

kenneth-bolivar-castro:
Added new UserLoginCleanAttemptsCommand. I added two @todo, please let me know what you think about them.


kenneth-bolivar-castro:
I refactored source code, now It's using a better way to retrieve current database connection.

PR #793, opened 2015-08-16
Rename EntityListController to EntityListBuilder.

pfrenssen:
In [Issue #2120871: Rename EntityListController to EntityListBuilder](https://www.drupal.org/node/2120871) the classes that generate the entity listings have been renamed and moved into the root namespace of the modules. We should do the same in our generated code.


pfrenssen:
This still needs to be tested.

jmolivas:
@pfrenssen: thanks for the PR, seems good to me, I will do some test of the code generation and reply with more comments.

jmolivas:
Found two bugs and creted one issue https://github.com/hechoendrupal/DrupalConsole/issues/808 the other was already reported on this issue https://github.com/hechoendrupal/DrupalConsole/issues/807

jmolivas:
I will merged this on and send a PR fixing the reported issues.

PR #798, opened 2015-08-17
797 generate entity content permission

jjcarrion:
With this PR we set the permission add/edit/delete to the entity created.


jmolivas:
@jjcarrion: Thanks for the PR

PR #803, opened 2015-08-20
Wrong handling of Entity::save() return value

pfrenssen:
Currently the entity form code that is generated thinks that the return value of `Entity::save() ` is a boolean indicating success or failure, but this is not correct. This probably changed at some point in the past.

In current HEAD if a failure occurs during the saving of the entity an exception is thrown. Instead of a boolean, this method returns either `SAVED_NEW` or `SAVED_UPDATED`. We can use this information to provide a more fitting message.

### DIFF
--- a/templates/module/src/Entity/Form/entity-content.php.twig
+++ b/templates/module/src/Entity/Form/entity-content.php.twig
@@ -57,15 +57,17 @@ class {{ entity_class }}Form extends ContentEntityForm {% endblock %}
     $entity = $this->entity;
     $status = $entity->save();
 
-    if ($status) {
-      drupal_set_message($this->t('Saved the %label {{ label }}.', array(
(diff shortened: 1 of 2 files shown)


pfrenssen:
Here is the relevant documentation from `EntityInterface`:

```
  /**
   * Saves an entity permanently.
   *
   * When saving existing entities, the entity is assumed to be complete,
   * partial updates of entities are not supported.
   *
   * @return int
   *   Either SAVED_NEW or SAVED_UPDATED, depending on the operation performed.
   *
   * @throws \Drupal\Core\Entity\EntityStorageException
   *   In case of failures an exception is thrown.
   */
  public function save();
```

jmolivas:
@pfrenssen: Thanks for the PR

PR #804, opened 2015-08-20
Use {@inheritdoc}

pfrenssen:
Current documentation standards suggest to use `{@inheritdoc}` when overriding parent methods.


jmolivas:

Issue #801, opened 2015-08-18
Invoke UserLoginCleanAttemptsCommand within UserPasswordResetCommand


kenneth-bolivar-castro:
After we reset user password we should clear all login failed attempts in order to be ready to start session without any issue related with account temporarily blocked.

jmolivas:
@kenneth-bolivar-castro: You can read about this feature at:
http://jmolivas.com/commands-calling-commands-in-drupal-console

And see some code examples at:
- GeneratorControllerCommand  
 ` $this->getHelper('chain')->addCommand('router:rebuild');`  https://github.com/hechoendrupal/DrupalConsole/blob/master/src/Command/GeneratorControllerCommand.php#L104

- SiteMaintenanceCommand  
  `$this->getHelper('chain')->addCommand('cache:rebuild', ['cache' => 'all']);`
 https://github.com/hechoendrupal/DrupalConsole/blob/master/src/Command/SiteMaintenanceCommand.php#L59

PR #811, opened 2015-08-23
#801 Added chain command to clear login failed attempts

kenneth-bolivar-castro:
Execute command UserLoginCleanAttemptsCommand after user account reset its password.


jmolivas:
@kenneth-bolivar-castro: Thanks for the PR this is how the execution looks like
pass-reset-clear-login

PR #827, opened 2015-08-30
Use routes array instead of method_name variable

jmolivas:


jmolivas:
This was broken because of an issue with box-project, fixed with https://github.com/box-project/box2/pull/107

Issue #299, opened 2015-01-13
[generate:entity:content] add bundles support


christianwiedemann:
Hi,
it would be nice to add a bundle option to the generate:entity:content command, which generates also the ConfigEntityType Entity (e.g. NodeType for Nodes) to have bundle support. 

Thanks 
Christian

socketwench:
Bundle generation would be a great addition since it's not obvious. Instead of making an option under one of the generate commands, it would make more sense to me for another top-level command: generate:entity:bundle?

jmolivas:
Like the idea of separating the command. I will look for the docs and paste it here to have a better idea of what is needed.

cosmicdreams:
I talked with @jmolivas about this issue.

Sounds like we have plan for implementing it.  The plan:

1. Create a content type and export it's config
2. Create a command to import the config while placeholdering / replacing the bundle name 
3. Assign the command to generate:content:type
4. Document it and SHIP IT

cosmicdreams:
I was able to complete step 1, but in looking at what I will need to do for step 2 I was wondering...  Where should I put the config I exported?  Is there another command I can look at for guidance on the kind of work I'll need to do in order to finish this?

cosmicdreams:
I have worked up the basics of what I mentioned above.  I don't seem to have the ability to push these changes to you because of permission issues.

PR #828, opened 2015-09-02
feature/DC299: Generate new Content type.

cosmicdreams:
- Add new generator and command to create new content types based on sample config.
- Solution for https://github.com/hechoendrupal/DrupalConsole/issues/299


src/Generator/GeneratorInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Drupal\AppConsole\Generator;
+
+interface GeneratorInterface {
+
+  public function generate($module, $class_name, $routes, $test, $services, $class_machine_name);
+}
```
heddn: Caught. Missing whitespace.

src/Generator/ContentTypeGenerator.php
```diff
@@ -0,0 +1,87 @@
...
+     * Generate node.type.{ bundle_name }.yml
+     */
+    $this->renderFile(
+      'module/src/Entity/Bundle/node.type.yml.twig',
+      $this->getSite()->getModulePath($module) . '/config/install/node.type.' . $bundleName . '.yml',
+      $parameters
+    );
+  }
+
+}
```
heddn: Caught again.

templates/module/src/Entity/Bundle/node.type.yml.twig
```diff
@@ -0,0 +1,17 @@
...
+    module:
+        - menu_ui
+third_party_settings:
+    menu_ui:
+        available_menus:
+            - main
+        parent: 'main:'
+name: {{ bundle_title }}
+type: {{ bundle_name }}
+description: 'Generated by Drupal Console'
```
heddn: Are quotes usually used in yml files? I seem to recall they weren't used. Please educate me.

seanr: They would indeed seem to be superfluous here, but not always - see http://stackoverflow.com/a/22235064/340462

Issue #879, opened 2015-09-18
[generate:module] should escape user-entered input (e.g. quotes)


anavarre:
If you type the below description:

```
$ console generate:module

 Welcome to the Drupal module generator 

Enter the new module name: The module I'd like to build
```
You'll get this in code:

```php
/**
 * Implements hook_help().
 */
function config_backup_help($route_name, RouteMatchInterface $route_match) {
  switch ($route_name) {
    // Main module help for the test module.
    case 'help.page.test':
      $output = '';
      $output .= '<h3>' . t('About') . '</h3>';
      $output .= '<p>' . t('The module I'd like to build.') . '</p>';
      return $output;

    default:
  }
}
```

Problem is the quote is not escaped, causing a PHP parse error when the module is enabled. It would be preferable to either escape the quotes or wrap strings in double quotes to prevent this.

jmolivas:
@anavarre: Nice catch, thanks for reporting, we will take care.

VasquezPonte:
https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/t/7

jmolivas:
Fixed with PR https://github.com/hechoendrupal/DrupalConsole/pull/940

PR #921, opened 2015-09-25
[generate:module] Wrap strings in double quotes.

VasquezPonte:
#879


jmolivas:
@VasquezPonte: Thanks for the PR
We can probably apply the `escape` twig filter to user-entered input
Instead of this
```
$output .= '<p>' . t("{{ description }}") . '</p>';
```
Change for this
```
$output .= '<p>' . t('{{ description|escape }}') . '</p>';
```

VasquezPonte:
I agree to it (see #940 ).

PR #932, opened 2015-09-28
Use <root>/autoload.php

webflo:
Console does not work with https://github.com/drupal-composer/drupal-project properly. Looks like refactoring to DrupalHelper introduced the bug.

Here is the failing Travis CI test. https://travis-ci.org/drupal-composer/drupal-project/jobs/81138968


jmolivas:
@webflo Thanks for the PR, I will test and merge.

PR #940, opened 2015-09-29
Escape user entered input using the twig filter.

VasquezPonte:
The TWIG escape filter is more suitable for the purpose.


jmolivas:
@VasquezPonte: Thanks for the PR.

jmolivas:
@VasquezPonte: Thanks for the PR

PR #946, opened 2015-09-30
Hande the changed field properly. Fixes #945

### LINKED ISSUE #945: [generate:entity:content] Hande the changed field properly
Right now Console generates content entities with the "changed" property and the getChangedTime()
getter that isn't on any interface.

What needs to be done:
1) The content entity interface needs to extend Drupal\Core\Entity\EntityChangedInterface.
2) The content entity class should use the Drupal\Core\Entity\EntityChangedTrait.
3) getChangedTime() should be removed from the class because it is provided by the trait.

bojanz:
As described in #945


jmolivas:
@bojanz: Thanks for the PR, I will review and merge.

PR #1019, opened 2015-10-09
[console] Detect root on Windows machines

tloeffler:
Added the detection for a root on a Windows drive.


jmolivas:
Thanks for the PR

PR #1025, opened 2015-10-09
[console] Supported Version outdated

stevenspasbo:
* Updated Supported Version in README and About command


jmolivas:
@stevenspasbo: Thanks for the PR

PR #1037, opened 2015-10-12
[console] Add _admin_route to admin routes.

joshuataylor:
https://www.drupal.org/node/2224207
* Routes starting with "/admin" will get by default TRUE value assigned.

Which means by default, people should be okay. This is up for discussion whether to include or not :)


jmolivas:
:+1: on this one.

PR #1038, opened 2015-10-12
[console] Entity access control handler no longer has a $langcode parameter

joshuataylor:
See https://www.drupal.org/node/2581447 , this results in an error when generating entities.


jmolivas:
@joshuataylor: Thanks for the PR

PR #1071, opened 2015-10-17
Add minor fixes to site:install command

favrik:
* Use correct translation keys for langcode and db-type
* Fix spacing of method comments
* Update code comments to use correct option


jmolivas:
@favrik: Thanks for the PR

PR #1072, opened 2015-10-17
[theme:download] Increase curl timeout to 10 seconds

favrik:
I was trying to download the zircon theme and was always getting timeout errors.  [Zircon](https://www.drupal.org/project/zircon) is about 2.66 Mb in size, which might not be an common occurrence.  Setting 10 seconds as the timeout was enough for me to be able to download it.

A couple of comments:
* Not sure if this is the proper fix, because network conditions vary across the world and some themes might be smaller or bigger. :/
* I noticed that `module:download` and `site:new` also create a `Buzz` browser (that uses curl as the client).  The construction/configuration of the `Buzz` client could be done, perhaps, in a helper.
* One more thing, the following code:
```php
   try {
            if (method_exists($client, 'get')) {
                $response = $client->get($release_file_path);
            } else {
```
will never execute, because the `Curl` client from `Buzz` doesn't have a `get()`  method.  Unless, I'm wrong. xD

### DIFF
--- a/src/Command/ThemeDownloadCommand.php
+++ b/src/Command/ThemeDownloadCommand.php
@@ -30,6 +30,7 @@ class ThemeDownloadCommand extends Command
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $client = new Curl();
+        $client->setTimeout(10);
         $browser = new Browser($client);
 
         $theme = $input->getArgument('theme');


favrik:
Update: I think this is not going to work for the long run.  Possible actions:
* Use drush to download stuff
* Use similar strategies as drush to download archives (git, wget, curl via `exec()`)
* Find another curl library more geared for downloading files
* Use curl via PHP with a "download large files" strategy

jmolivas:
@favrik: I will merge this and add the same `$client->setTimeout(10);` temporary fix to `site:new` and `module:download` commands.

I agree on the idea of a helper. I will add the helper class and move that functionality to it.

Issue #767, opened 2015-08-02
Console don't running at beta 13


vacho:
This error appear.. 

drupal generate:module
PHP Fatal error:  Cannot redeclare GuzzleHttp\Psr7\str() (previously declared in phar:///usr/local/bin/drupal/vendor/guzzlehttp/psr7/src/functions.php:17) in /var/www/html/devbeta13/core/vendor/guzzlehttp/psr7/src/functions.php on line 39

alexander-bogomolov:
Issue from Drupal.org: https://www.drupal.org/node/2538484

jmolivas:
@vacho / @alexander-bogomolov 
We just tagged a new release 0.8.0 please update the project, for more information about updating please read:   
- http://hechoendrupal.gitbooks.io/drupal-console/content/using/update.html  
- http://hechoendrupal.gitbooks.io/drupal-console/content/contributing/keeping-your-fork-up-to-date.html

jmolivas:
Fixed with PR https://github.com/hechoendrupal/DrupalConsole/pull/771 merged on this release https://github.com/hechoendrupal/DrupalConsole/releases/tag/0.8.0

alexander-bogomolov:
Thank you!

PR #1077, opened 2015-10-19
[helper] Re-add Guzzle for handling file downloads

favrik:
This PR adds the following changes:
* Remove Buzz
* Add Guzzle ~6.1
* Add HttpClientHelper to re-use download logic
* Update theme:download, site:new, and module:download commands

As described on #1072, downloading stuff is unreliable due to timeouts.  Apparently, `Buzz` does not allow an empty timeout setting (see [AbstractClient](https://github.com/kriswallsmith/Buzz/blob/master/lib/Buzz/Client/AbstractClient.php#L9) and [AbstractCurl](https://github.com/kriswallsmith/Buzz/blob/master/lib/Buzz/Client/AbstractCurl.php#L210-L215)), but I might be totally wrong.  If that's the case, please let me know.

So, I began testing with latest `Guzzle` successfully.  However, `Guzzle` was used previously on the project, but due to #767 it was replaced by `Buzz`.

I tried to replicate #767, but was unable to, maybe I'm using an incorrect box build command.  My current command is simply `box build -v` inside the project directory.

Thoughts on course of action?  Can you provide the current box build command? Can you help test that this does not introduce a regression?  Thanks!!

### DIFF
--- a/bin/console.php
+++ b/bin/console.php
@@ -27,6 +27,7 @@ use Drupal\Console\EventSubscriber\ShowGenerateDocListener;
 use Drupal\Console\Helper\DrupalHelper;
 use Drupal\Console\Helper\CommandDiscoveryHelper;
 use Drupal\Console\Helper\RemoteHelper;
+use Drupal\Console\Helper\HttpClientHelper;
 
 set_time_limit(0);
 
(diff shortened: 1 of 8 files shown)


jmolivas:
@favrik: Thanks for the PR. 

I tested this and it works fine.

The changes here https://github.com/guzzle/guzzle/blob/master/src/functions_include.php reflect the fixes to the previous issue.

We are good to go. I will merge this and include on the next release.

PR #1084, opened 2015-10-21
[site:status] Check for translatable strings before calling render()

### LINKED ISSUE #1082: [site:status] PHP Fatal error:  Call to a member function render() on string in src/Command/SiteStatusCommand.php on line 94
PHP version: 5.6
Drupal Console version: 0.9.4
Drupal version: 8.0.0-rc1 

Getting the following fatal error when executing

**drupal site:status**

PHP Fatal error:  Call to a member function render() on string in src/Command/SiteStatusCommand.php on line 94

favrik:
The `locale` module does not use `TranslatableMarkup` strings for the title of its requirements, I'm pretty sure this is by design since its status can't depend on a locale being available or completely translated.

Fixes #1082

### DIFF
--- a/src/Command/SiteStatusCommand.php
+++ b/src/Command/SiteStatusCommand.php
@@ -91,9 +91,13 @@ class SiteStatusCommand extends ContainerAwareCommand
         $systemData = [];
 
         foreach ($requirements as $key => $requirement) {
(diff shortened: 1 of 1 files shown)


jmolivas:
@favrik: Thanks for the PR

PR #1090, opened 2015-10-22
[event:subscriber] Fix generator

### LINKED ISSUE #1089: generate:event:subscriber is in very bad shape
Here's what the command gave me:

```php
<?php

/**
 * @file
 * Contains Drupal\broken_form\DisplayMessageSubscriber.
 */

namespace Drupal\broken_form\EventSubscriber;

use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\HttpKernel\Event\GetResponseEvent;

/**
 * Class DisplayMessageSubscriber.
 *
 * @package Drupal\broken_form
 */
class DisplayMessageSubscriber implements EventSubscriberInterface {
  /**
   * Constructor.
   */
  public function __construct() {

  }
/**
* Registers the methods in this class that should be listeners.
*
* @return array
*   An array of event listener definitions.
*/
static function getSubscribedEvents() {

  $events['config.save'] = array('displayMessage');
  return $events;
}

/**
 * This method is called whenever the config.save event is
 * dispatched.
 *
 * @param GetResponseEvent $event
 */
 public function displayMessage(GetResponseEvent $event) {
   drupal_set_message('Event config.save thrown by Subscriber in module broken_form.', 'status', TRUE);
 }

}
```

We need to fix the indentation, and the usage of old array syntax while we're at it.
getSubscribedEvents() should use inheritdoc since the method is defined by the interface.
And typehinting GetResponseEvent will crash most event subscribers, it's just the wrong class.

bojanz:
Fixes #1089 (mostly).

### DIFF
--- a/templates/module/src/event-subscriber.php.twig
+++ b/templates/module/src/event-subscriber.php.twig
@@ -10,7 +10,7 @@ namespace Drupal\{{module}}\EventSubscriber;
 
 {% block use_class %}
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
-use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\EventDispatcher\Event;
 {% endblock %}
 
 {% block class_declaration %}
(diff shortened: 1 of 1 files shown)


enzolutions:
Thanks @bojanz for your contribution

PR #1102, opened 2015-10-24
Add more complete instructions for Bash completion

kepford:
SInce completion in Bash depends on [bash-completion](http://bash-completion.alioth.debian.org/) it should be noted.


omero:
Thanks for the PR @kepford :)

PR #1103, opened 2015-10-24
Updating line width in BASH Completion instructions of README

mark-casias:
Just so it doesn't scroll left forever.


jmolivas:
@mark-casias: Thanks for the PR

PR #1107, opened 2015-10-26
[translate] Add Samuel Jackson Language 

jmolivas:


darol100:
Closing this should not be part of DC core.

PR #1109, opened 2015-10-26
[translation] Add English and Spanish translation for the generate:entity:config command

dinarcon:


jmolivas:
@dinarcon: Thanks for the PR

Issue #1095, opened 2015-10-23
Aliases Discussion


darol100:
## Problem / Motivation
We should provide aliases (shortcuts) to simplify the Drupal Console commands.

## Solution
We should provide a list of recommend aliases and discuss them.

dmouse:
You can execute each command only with the initial(s) letters, for example

console g:c
console gen:cont

darol100:
@dmouse but we should provide a list of different types commands that we can improve. g:c is a good start but we can create more aliases.

jmolivas:
The idea of this issue, is to specify the aliases instead of letting the component suggest.

jmolivas:
I sent a PR https://github.com/hechoendrupal/DrupalConsole/pull/1128 to add command alias via user configuration file `~/.console/config.yml` 

* Configuration of command alias: https://github.com/jmolivas/DrupalAppConsole/blob/1095-console-command-alias-implementation/config/dist/config.yml#L26-L29

* Discovery of aliases: https://github.com/jmolivas/DrupalAppConsole/blob/1095-console-command-alias-implementation/src/Application.php#L246

* Setting alias to command: https://github.com/jmolivas/DrupalAppConsole/blob/1095-console-command-alias-implementation/src/Application.php#L259-L267

I know this issue is about defining the aliases to be assigned to commands, but I wanted to make this feature based on a configuration file to avoid changing code and make easy to override by the user.

jmolivas:
After a reviewing the PR https://github.com/hechoendrupal/DrupalConsole/pull/1128 the following changes:
- Define aliases on a separated file `aliases.yml` instead of `config.yml`
- Use this YAML structure:
```
application:
  aliases:
    commands:
      cache:
        rebuild:
            - cr
```

jmolivas:
PR https://github.com/hechoendrupal/DrupalConsole/pull/1128 was merged 

We should start a discussion about best alias names for commands we will use,

These alias will be a recommendation and user can override by updating the copied file `~/.console/aliases.yml`

darol100:
@jmolivas is this implementation going to have the aliases display in the drupal list  command?

For instance `drupal cache:build (cr)`. This will make user to easily learn the aliases.

darol100:
Here are my suggestion for the new aliases.


### I do not think we need aliases for these
_completion
about
chain
complete
drush
help
init
list
self-update
              
 Cache
==========================      

  cache:rebuild                    = cr

 Config
[...]

PR #1128, opened 2015-10-31
[console] Add command alias registration #1095

jmolivas:


jmolivas:
This PR add command aliases via user configuration file `~/.console/config.yml` 

* Configuration of command alias: https://github.com/jmolivas/DrupalAppConsole/blob/1095-console-command-alias-implementation/config/dist/config.yml#L26-L29

* Discovery of aliases: 
 - https://github.com/jmolivas/DrupalAppConsole/blob/1095-console-command-alias-implementation/src/Application.php#L246
 -  https://github.com/jmolivas/DrupalAppConsole/blob/1095-console-command-alias-implementation/src/Application.php#L259-L267

* Setting alias to command: https://github.com/jmolivas/DrupalAppConsole/blob/1095-console-command-alias-implementation/src/Application.php#L248

I know this issue is about defining the aliases to be assigned to commands, but I wanted to make this feature based on a configuration file to avoid changing code and make easy to override by the user.

jmolivas:
After review with @omero and @enzolutions these are the proposed changes:
- Define aliases on a separated file aliases.yml instead of config.yml
- Use this YAML structure:
```
application:
  aliases"
      cache:
        rebuild:
          aliases:
            - cc
```

PR #1138, opened 2015-11-01
[translate] Hindi generate.*

dkgndec:
* generate.controller.yml
* generate.entity.bundle.yml
* generate.event.subscriber.yml
* generate.form.alter.yml
* generate.form.yml


enzolutions:
Thanks @danielnv18 @subhojit777  and others folks

dkgndec:
It should be @dkgndec

PR #1153, opened 2015-11-06
Fixes the currently broken get of existing entity-langcode

tsvenson:
The current code for getting the entity langcode is broken, I'm not entirely sure this is the correct way of doing this but it does work when editing existing entities.

I also found https://www.drupal.org/node/2303877 which hints using "language()->getId()" should be avoided for content entities.


jmolivas:
@tsvenson : Thanks for the PR. We will test this and merge this.

PR #1159, opened 2015-11-07
[generate:doc] Rename the current generate:doc and add a command for Dash

FatherShawn:
We would like to have Drupal Console documentation available by through the [Dash application](https://github.com/Kapeli/Dash-User-Contributions#contribute-a-new-docset).

Refactored generate:docs into generate:docs:gitbook and added generate:docs:dash which creates a Dash docset.


jmolivas:
@FatherShawn: Thanks for the PR. We will test this and merge this.

enzolutions:
Hi @FatherShawn 

I tested and it's almost done. I imported the docset in my Dash and I can read the documention and follow the links.

But, there is always a but, I can't follow the index as you ca see in the following image.

console_dash_index_error

I am also wondering if you can add some level of css to tidy the documentation in dash?

FatherShawn:
OK - the path is clearly not happy in that section.

I'm not much of a designer, but I'll make a stab at some css

FatherShawn:
@enzolutions OK - path fixed.  I'll see about some CSS

jmolivas:
If links are fixed we can go as it is and ask for help to someone related to design/css.

FatherShawn:
I added some text styling so it is not quite so raw.

templates/dash/index.html.twig
```diff
@@ -0,0 +1,153 @@
...
+    <table>
+        <thead>
+        <tr>
+            <th>Option</th>
+            <th>Details</th>
+        </tr>
+        </thead>
+        <tbody>
+        {% for option in options %}
+            <tr class="{{ cycle(['odd', 'even'], loop.index0) }}">
```
FatherShawn: This twig cycle is not adding classes like it does on the `generate-doc.html.twig` template but I don't understand why.

enzolutions:
Great @FatherShawn I will merge the command to include in next release, could you help us in the process to register this docset as part of Dash?

FatherShawn:
Sure - I'll check it out

src/Command/GenerateDashCommand.php
```diff
@@ -0,0 +1,241 @@
...
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Exception\IOExceptionInterface;
+
+class GenerateDashCommand extends ContainerAwareCommand
+{
+
+    /**
+     * @constant Contents of the plist file required by the docset format.
+     */
+    const PLIST = <<<PLIST
```
dmouse: Maybe we can put this ```const``` into a twig file like the html code? cc @FatherShawn @enzolutions

PR #1184, opened 2015-11-11
[travis] Adding site commands to use on travis

omero:


jmolivas:
@omero we should add support to `sqlite` on the `site:install` command
https://github.com/hechoendrupal/DrupalConsole/blob/master/.travis.yml#L34

A temporary solution could be to add `mysql` to `.travis.yml` config file to use `site:install` command.

omero:
Yes @jmolivas  i know, but that can slow the build, let me see if i can obtain the `sqlite` support for `site:install` command

PR #1223, opened 2015-11-13
[translations] English yaml.merge

alex-arriaga:


jmolivas:
@alex-arriaga: single quote scape http://yaml.org/spec/current.html#id2534365

PR #1259, opened 2015-11-15
[docs] Add licence file and badge for it

karmazzin:
I believe we need to add licence file to repository, it is a good practice. As I understand, DrupalConsole uses MIT licence. I prepared licence file and badge for it.


jmolivas:
@karmazzin: Thanks for the PR

Issue #1286, opened 2015-11-17
[docs] Theme debugging in Drupal.org


darol100:
## Problem/Motivation
Drupal.org have documentation about debugging twig theme at [Debugging compiled Twig templates](https://www.drupal.org/node/1903374). However, this seem like a lot steps to do manually and in theory we should be able to run `drupal site:mode dev`

## Solution

We should verify that everything on that page is supported [Debugging compiled Twig templates](https://www.drupal.org/node/1903374) by `drupal site:mode dev`. I know that debug true is been supported already, in addition @jmolivas confirm it at https://github.com/hechoendrupal/DrupalConsole/issues/1148

````````````````
parameters:
  twig.config:
    debug: true
````````````````````
We need to check if auto_reload, disable cache is also been supported. If not been supported we should open issue couple issues to support it. Here is how it should looks like.

```````
auto_reload: true
cache: false
```````

Last thing we need to check if 
[Drupal Render cache](https://www.drupal.org/node/1903374#drupal-render-cache ) is been support it by  `drupal site:mode dev`

Once we verify that all that is been supported by `drupal site:mode dev` we can added a piece a documentation that mention if you are running Drupal Console run `drupal site:mode dev` and you have debugging enable. 

## Why ? 

We need to start documenting DrupalConsole functionality, and providing solutions to developers so they can use Drupal Console as part of their workflow.

**Comment by jmolivas** [comment]
Those values are switch by `drupal site:mode dev`

```
twig.config: { debug: true, auto_reload: true, cache: false }
```
![drupal-site-mode](https://cloud.githubusercontent.com/assets/366275/11223187/89edca58-8d23-11e5-99fb-bb0a1b5aebbe.png)

Totally agree on the need to to start documenting DrupalConsole functionality

darol100:
@jmolivas, what about the [Drupal Render cache](https://www.drupal.org/node/1903374#drupal-render-cache) ? Its that been supported as well ? If not, should we open an issue for that ?

jmolivas:
@darol100 Drupal Render cache toggle not yet supported, and yes we should open en issue for that.

darol100:
@jmolivas , I created a new issue for the Drupal Render Cache.  #1309

I have also update the documentation in Drupal.org on how to enable debugging with Drupal Console. You can see the [revision in here](https://www.drupal.org/node/1903374/revisions/view/9122064/9128298). 

[...]

PR #1310, opened 2015-11-20
[site:mode] Do not disable Twig cache in dev mode

Cottser:
## Problem/Motivation

When you use debug (or just auto_reload if you for some reason don't want debug on) the Twig cache doesn't get in your way. Disabling the Twig cache just makes for a slower development experience because each template needs to be compiled regardless of if it's been edited or not. Also, you can't easily debug/look at the compiled Twig templates (PHP classes) if they are not cached.

## How to reproduce

n/a I think, this is not a bug report.

## Solution

Leave Twig cache alone! I think 99.9999% of the time developers should only touch debug, and leave auto_reload and cache alone. I would argue that auto_reload can be left as `null` but don't want to scope creep :)

### DIFF
--- a/src/Command/Site/ModeCommand.php
+++ b/src/Command/Site/ModeCommand.php
@@ -201,7 +201,7 @@ class ModeCommand extends ContainerAwareCommand
             'twig.config' => [
                 'debug' => ['dev' => true, 'prod' => false],
                 'auto_reload' =>['dev' => true, 'prod' => false],
-                'cache' => ['dev' => false, 'prod' => true]
+                'cache' => ['dev' => true, 'prod' => true]
             ]
         ];
 


omero:
The goal of this command is improve the `dev` experience so ... :+1: Thanks for the PR @Cottser

jmolivas:
Thanks for the PR @Cottser  :+1: and thanks for review and merge @omero

Cottser:
Thanks :)

darol100:
@Cottser , Probably this is not relate to this issue but... Do you think we should explain why not to disable the cache in https://www.drupal.org/node/1903374 ? On a previous issue https://github.com/hechoendrupal/DrupalConsole/issues/1286 we was trying to make sure that DrupalConsole matches their mode: dev with the Drupal.org documentation. I think if we do not explain that disabling the cache is not good all the time is going to looks like DrupalConsole does not support that feature.

@jmolivas Should we open an issue separate issue to support this separately? Maybe not part of site:mode dev but if this feature is part of D8, I will assume that is because sometimes this might be useful even thought is only 00.0001% ?

jmolivas:
@Cottser: @darol100: 
I sent a PR https://github.com/hechoendrupal/DrupalConsole/pull/1316 related to make this values customizable. 

Configuration is stored at: https://github.com/hechoendrupal/DrupalConsole/blob/master/config/dist/site.mode.yml

The `site.mode.yml` file must be copied to your home directory by executing the `init` command.

Cottser:
@darol100 I updated the docs at https://www.drupal.org/node/1903374, if you or someone can please update the embedded screenshot that'd be awesome :) thanks!

PR #1325, opened 2015-11-23
Add exit code for Application::doRun()

zviryatko:
Symfony console application expected that method doRun() will return an exit code: zero if everything went fine, or an error code.
It prevents `The command terminated with an error status ()` error after every command in console.


jmolivas:
@zviryatko Thanks for the PR. Nice catch.

PR #1335, opened 2015-11-24
[translation] Vietnamese

webtomme:
Add vietnamese translations.
Some statement is not translated, they contains some very IT specific terms which do not have similar vietnamese words or very confusing to use these words in vietnamese.


enzolutions:
Resolved conflict and create new PR https://github.com/hechoendrupal/DrupalConsole/pull/1342

PR #1341, opened 2015-11-24
[template] Fix deprecated function usage

zviryatko:
Method EntityListBuilder::getLabel() is deprecated, use $entity->label() instead.

Link to the docs:  
https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Entity!EntityListBuilder.php/function/EntityListBuilder%3A%3AgetLabel/8


jmolivas:
@zviryatko: Thanks for the PR.

PR #1343, opened 2015-11-24
[console] Added validator to String helper to check machinename length

enzolutions:


nikathone:
The proposed solution will definitely fix problem but the user is unaware that his machine name will be truncated. If we can let the user know that the string was truncated that will be helpful.

Keep up the good work. Thanks

jmolivas:
@nikathone true we should send a message notifying the user, something like: 

> The max length supported is 32, the your entered value was fixed as NEW_VALUE.

nikathone:
@jmolivas yeah that will do the trick.

PR #1354, opened 2015-11-25
[WIP] - [server] New command

enzolutions:


jmolivas:
@enzolutions Should we keep working on this?

We already have the server command starting the PHP built-in server.

PR #1357, opened 2015-11-26
Use Psr/Log

greg-1-anderson:
Make MessageHelper implement LoggerInterface, so that it may be provided to any class that needs a Prs\Log-compatible logger.  The purpose here is to help enable code sharing, although at this particular time there is no candidate code to share.

I do find it somewhat troubling that Symfony/Console does not use Psr/Log, but since DrupalConsole already happens to have a MessageHelper, this extension is somewhat natural.

### DIFF
--- a/src/Helper/MessageHelper.php
+++ b/src/Helper/MessageHelper.php
@@ -7,22 +7,27 @@
 
 namespace Drupal\Console\Helper;
 
+use \Psr\Log\LoggerInterface;
+use \Psr\Log\LoggerTrait;
+use \Psr\Log\LogLevel;
 use Drupal\Console\Helper\Helper;
 
-class MessageHelper extends Helper
+class MessageHelper extends Helper implements LoggerInterface
 {
+    use LoggerTrait;
+
     /**
      * @var string
      */
-    const MESSAGE_ERROR = 'error';
+    const MESSAGE_ERROR = LogLevel::ERROR;
     /**
      * @var string
      */
-    const MESSAGE_WARNING = 'warning';
+    const MESSAGE_WARNING = LogLevel::WARNING;
     /**
      * @var string
      */
-    const MESSAGE_INFO = 'info';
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
c.f. https://github.com/drush-ops/drush/pull/1808

The general direction in Drush is to consider using [Robo](http://robo.li/) as a more object-oriented, less-code-to-maintain way to implement a number of operations. Robo has a cli parser, but I am only proposing using its class libraries. The irony here is that Robo does not use Psr/Log, but we can always submit a PR there and see what happens. As it is, I/O is a roadblock to use of these libraries; we'd need to solve that one way or another.

Don't know if any of this is of any interest, just thought I'd throw this in as a suggestion.

jmolivas:
Thanks @greg-1-anderson, looks good I will take a look and totally agree on implementing this.

jmolivas:
@greg-1-anderson Maybe another irony is that robo is using the symfony/console component https://github.com/Codegyre/Robo/blob/master/composer.json#L20 as we do j/k

greg-1-anderson:
It sort of follows, that if symfony/console does not use Psr/Log, then by default, the components that implement it will not either.

I have an idea for fixing this problem in the upstream component. I will submit a PR in symfony console, and reference it here when I do.

jmolivas:
I keep looking at this.

greg-1-anderson:
I think it's fine to just merge this, as you have done. It will be a while before we have Psr/Log in Symfony/Console.  Using Psr/Log like this will allow progress to be made.  Thanks.

jmolivas:
I did to standard our messages class I will do some changes in the class to take advantage of `log` method and remove some methods at the class.

greg-1-anderson:
The main benefit will be, if there is some separate (or separable) library that needs to log, it should be initialized with the LoggerInterface from your Message class, rather than using the Message class directly. This is only important for parts of the code that might become separate libraries someday.

PR #1378, opened 2015-11-29
[self-update] Remove abandonded dependency fix #806

### LINKED ISSUE #806: Abandoned dependency? -- herrera-io/phar-update
It appears the dependency `"herrera-io/phar-update": "1.*"` is abandoned upstream per its' source GitHub owner?: https://github.com/kherge-abandoned/php-phar-update

Are there any plans to replace this dependency with a non-abandoned alternative?  I ask because I would like to eventually package this project downstream for Fedora/EPEL but I do not want to be responsible for packaging an abandoned dependency.

omero:
Replace abandoned dependency `herrera-io/phar-update` with `padraic/phar-updater` this fix #806

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,12 +50,12 @@
         "symfony/translation": "2.7.*",
         "symfony/yaml": "2.7.*",
         "twig/twig": "^1.22.2",
-        "herrera-io/phar-update": "1.*",
         "symfony/dom-crawler": "2.7.*",
         "alchemy/zippy": "0.2.*@dev",
         "phpseclib/phpseclib": "2.*",
         "stecman/symfony-console-completion": "^0.5.1",
-        "guzzlehttp/guzzle": "~6.1"
+        "guzzlehttp/guzzle": "~6.1",
+        "padraic/phar-updater": "~1.0@dev"
     },
     "bin": ["bin/console"],
     "config": {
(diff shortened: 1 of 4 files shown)


jmolivas:
This fixes #806

pjcdawkins:
As far as I can tell, this means that you no longer verify the SHA1 hash of the downloaded file. For the Platform.sh CLI I'm writing a custom 'strategy' for this library, which uses the old (herrera-io) style "manifest.json" to contain the versions and SHA1 hashes:
https://github.com/pjcdawkins/platformsh-cli/blob/replace-phar-update/src/SelfUpdate/ManifestStrategy.php

jmolivas:
@pjcdawkins Yes we skip the validation and planned to write it later, but we can probably re-use what you did ?

Any reason why you wrote a custom 'strategy' instead of using the one provided by the library https://github.com/padraic/phar-updater#basic-sha-1-strategy ?

pjcdawkins:
The one provided by the library requires you to have the SHA1 in a file on its own at URL - it can't look through a list of versions. Also I included a little bit of logic for the "Lock to major version" option that herrera-io/phar-update had (I've named it as a parameter `$allowMajor`, which you'd set true to allow updates between major versions).

PR #1388, opened 2015-11-30
Fix path for LICENSE badge

karmazzina:


jmolivas:
@karmazzina: Thanks for the PR and good catch.

PR #1393, opened 2015-11-30
#1391: declare additionnal entity keys, Fix #1391

### LINKED ISSUE #1391: [generate:entity:content] Complete entity declaration against table
Hi

`entity_keys` missed some entries : 
* "uid" => "user_id"
* "langcode" = "langcode"

by the fact there is a column named "langcode", I think it should be necessary to add "translatable = TRUE" in the entity declaration itself.

reading similarities with node declaration, I see "common_reference_target = TRUE".
I don't know exactly what does it mean, but I suggest to add it too.

smalot:
PR for #1391 issue,
Hope it helps


jmolivas:
@smalot: Thanks for the PR.

Issue #1399, opened 2015-12-02
[database:log:debug] could do with a --reverse


cburschka:
There's a `--limit` and an `--offset`, but since those are applied in ascending order, the only way to get the most recent events is to let it print everything, and pipe it into `tail`.

Having a `--reverse` option that adds an `orderBy('wid', 'DESC')` in there might save some time with large logs.

jmolivas:
:+1: on this one.

PR #1400, opened 2015-12-02
#1399 Add a --reverse option to database:log:debug

cburschka:
Implementing #1399.


jmolivas:
@cburschka: Thanks for the PR

Issue #295, opened 2015-01-11
[generate:profile-stub] Add support for generating profile stubs


jmolivas:
Moved from URL at d.o. https://www.drupal.org/node/2392825 created by Sam152 
I think generating stubs for profiles would be great. It would also be really handy to 
have the generated profile be based on one of the core profiles, basically allowing 
you to use standard as a starting point.

```
jmolivas commented 22 days ago:
@Sam152 may you share the documentation link ?, for us to take a look and find out 
is this feature is doable

Sam152 commented 22 days ago:
Here is the profile documentation: https://www.drupal.org/node/2210443

Most profile are forked from "standard" or "minimal" which include quite a bit of 
boilerplate. It might be beyond the scope of this tool to provide a profile based on one 
of these starting points, as there are no special allowances to allow fork-ability, 
but it would be nice.

jmolivas commented 6 days ago:
We achieve an important step in order to be able to work on this issue, to have the 
global installer http://jmolivas.com/install-drupal-console-globally,

We will take a look at the doc link and your blog post 
https://codedrop.com.au/blog/test-driving-drupal-8-writing-install-profile in order to 
work and have a patch/PR for this issue.
```

pfrenssen:
This has been fixed in PR #1406. This can be closed.

jmolivas:
@pfrenssen: Thanks.

PR #1406, opened 2015-12-02
Add a generator for installation profiles.

pfrenssen:
This is a PR for issue #295.

### DIFF
--- a/Test/DataProvider/ProfileDataProviderTrait.php
+++ b/Test/DataProvider/ProfileDataProviderTrait.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Drupal\Console\Test\DataProvider;
+
+/**
+ * Class ProfileDataProviderTrait
+ * @package Drupal\Console\Test\DataProvider
+ */
+trait ProfileDataProviderTrait
+{
+    /**
+     * @return array
+     */
+    public function commandData()
+    {
+        $this->setUpTemporalDirectory();
+
+        return [
+          // Profile name, machine name, description, core version,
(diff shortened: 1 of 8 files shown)


pfrenssen:
This is based on the example at the official documentation: https://www.drupal.org/node/2210443. All code has been stolen from the module generator.

It's still in progress, I currently have an issue where Twig cannot find the template:

```
 [Twig_Error_Loader]                                                                                                 
  Unable to find template "profile/info.yml.twig" (looked into: /home/pieter/v/joinup/vendor/drupal/console/src/Command/../../templates).
```

pfrenssen:
Ok so the Twig issue was a bit special. I have DrupalConsole installed globally on my system and always use this global instance. I wasn't aware that the Drupal 8 site I was using to test also had DrupalConsole installed as a composer dependency. Twig was looking for the templates in the local composer managed instance instead of in the global instance I was running. This is maybe a bug, but unrelated to this PR.

I found a few small bugs when I was able to run the code and fixed them.

pfrenssen:
Added a test. I found a violation against proper OO decoupling thanks to the test :)

pfrenssen:
Note that the test is only checking that the files are being generated, it doesn't check the content of the files. This is in line with what the other tests are doing. I can extend the tests to check the contents as well if this is desired.

jmolivas:
@pfrenssen: At this point asking for the generated files is fine. we can decide later on if we want to validate file contents.

jmolivas:
@pfrenssen: Thanks for the  PR.

PR #1418, opened 2015-12-05
[translation] zh_hans

ranqiangjun:
1. Some translations of zh_hans
2. Deleted config/translations/*/db.client.yml as config/translations/en/db.client.yml was removed.
3. Added config/translations/*/generate.profile.yml as config/translations/en/generate.profile.yml was added.


enzolutions:
Thanks @ranqiangjun

PR #1435, opened 2015-12-06
Fix command theme:download

ranqiangjun:
Wrong description and options.


jmolivas:
@rachellawson: Thanks fir the PR.

PR #1444, opened 2015-12-07
[Console] Updated quick start chain to use Drupal 8.0.1.

danielnv18:
The command ```drupal init --override``` is setting in quick-start chain the Drupal version 8.0.0 when the 8.0.1 is already available


enzolutions:
Thanks @danielnv18

Issue #1423, opened 2015-12-05
hook_theme command


itsdarrylnorris:
### Problem / Motivation
#1422 I propose to remove hook_theme from `generate:module` because is an empty function. By removing this from the `generate:module` we need to add the ability of the user to generate hook_theme.
### Solution
- We need to be able to generate hook_theme on a stand a lone command.
- If there is an exiting hook_theme, this command should hook_theme instead of generating another hook_theme. Otherwise we will be running to the same issue as #1562 .

enzolutions:
Closing this issue, because was fixed and recently tested with issue https://github.com/hechoendrupal/drupal-console/issues/1562

PR #1466, opened 2015-12-14
Fixes #1422: Remove the empty hook_theme from generate:module command.

### LINKED ISSUE #1422: Remove the empty hook_theme from generate:module command
### Problem / Motivation
There is no need to have an empty hook_theme without any functionality when the user generate a module using the command `generate:module`.

``````
/**
 * Implements hook_theme().
 */
function test_3_module_theme() {
  $theme = [];

  return $theme;
}
```````

### Solution
Lets remove this unnecessary code.

darol100:
There is no need to have an empty hook_theme without any functionality when the user generate a module using the command generate:module.

### DIFF
--- a/templates/module/module.twig
+++ b/templates/module/module.twig
@@ -23,12 +23,4 @@ function {{machine_name}}_help($route_name, RouteMatchInterface $route_match) {
   }
 }
 
-/**
- * Implements hook_theme().
- */
-function {{machine_name}}_theme() {
-  $theme = [];
-
-  return $theme;
-}
 {% endblock %}


jmolivas:
We can probably add a new option `--hook-theme` andva question `Add hook theme?` or add a new command to add this separately.

darol100:
I was thinking solving this in a separate issue at https://github.com/hechoendrupal/DrupalConsole/issues/1423. Should we solve it as part of the PR as well ?

darol100:
@jmolivas anything missing from this PR ? Do you still want to support the hook_theme command as part of this story ?

If not, lets remove the hook_theme from the generate:module since is causing problems at #1562 .

jmolivas:
@darol100 Thanks for the PR.

PR #1467, opened 2015-12-14
[aliases] theme download fix

darol100:
This should said "theme" instead of themem. For this reason this will never run because there is not a such as thing as "themem"


enzolutions:
Thanks @darol100

PR #1468, opened 2015-12-14
Introducing aliases to all commands.

darol100:
This is just a start there still some work for this PR to be complete. 

 - We should add the ability of having the command aliases in the drupal list. For instance `drupal cache:build (cr)`. This will make user to easily learn the aliases.
 - We should investigate and how to add a priority, to the aliases over the auto detection of command. 
 - We should test all the commands to make sure that they are running property.

### DIFF
--- a/config/dist/aliases.yml
+++ b/config/dist/aliases.yml
@@ -4,9 +4,270 @@ application:
       cache:
         rebuild:
             - cr
-      themem:
+      theme:
         download:
-            - td
+           - td
       module:
         download:
             - md
+      config:
+        debug:
+          - cde
+      config:
+        edit:
+          - cdit
+      config:
+        export:
+           - ce
+      config:
+        export:
+          content:
+            type:
+              - cect
+      config:
+        export:
+          single:
+            - ces
+      config:
+        export:
+          view:
+            - cev
+      config:
+        import:
+          - ci
+      config:
+        import:
+          single:
+            - cis
+      config:
+        override:
+          - co
+      container:
+        debug:
+          - cod
+      cron:
+        execute:
+          - cre
+      cron:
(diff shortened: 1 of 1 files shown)


darol100:
@jmolivas some of the commands does not work, meaning that does not seem to be found. Some of the command conflict with the auto detection functionality.

And I do not think that when there is 3 words will not work. I'm not 10% sure. 

For instance 

      generate:
        plugin:
          fieldtype:
            - gpft

This will not work because have 3 words of indentation. Or at least does not work on my local. 

I will appreciate if you can provide some of feedback why you think some of these aliases does not work.

jmolivas:
@darol100 I tested as you mentioned some alias are not registered. 

I saw the `uslca` registered, the 3 nested levels on the command name may is not the issue. I will keep investigation on this and ping you when found the reason this is not working as expected.
```
      user:
        login:
         clear:
           attempts:
            - uslca
```

jmolivas:
I found the issue, is related to the YAML indentation and how the levels are nested.

I will merge this and fix it on my local

seanr:
I could be reading it wrong, but it looks like there are still a number of lines that are inconsistently indented in the commit.  Will that cause problems or is it purely cosmetic?

jmolivas:
@seanr: that is pure cosmetic the issue is:

```
      config:
        debug:
          - cde
      config:
        edit:
          - cdit
```
must be:
```
      config:
        debug:
          - cde
        edit:
          - cdit
```

jmolivas:
Fixed with this PR #1490

This was the final result of the `aliases.yml` file: https://github.com/hechoendrupal/DrupalConsole/blob/master/config/dist/aliases.yml

PR #1469, opened 2015-12-14
Hungarian for yaml.update

rmelike:


jmolivas:
@rmelike: Thanks for the PR.

Issue #1465, opened 2015-12-13
[generate:entity:content]  \Drupal:l() is deprecated.


martinfrances107:
\Drupal:l() is deprecated. 
In code I scaffold I would like to avoid this global function call.
The generator command is good .... so I want to fix the issue at source.

I will post a patch soon.

darol100:
See more about the deprecation at -https://api.drupal.org/api/drupal/core%21lib%21Drupal.php/function/Drupal%3A%3Al/8

martinfrances107:
I have made the changes in the commit above ... Sorry I don't use github frequently .....would it be better  to  produce a pull request?

darol100:
I create the PR for you - https://github.com/hechoendrupal/DrupalConsole/pull/1471. In addition, I search for `\Drupal::l` in the DrupalConsole project and here is the only part where we are using it.

martinfrances107:
@darol100 thank you.

jmolivas:
PR #1471 merged

PR #1471, opened 2015-12-14
Fixes for #1465

darol100:
Deprecate function `\Drupal::l` see more about the deprecation at -  https://api.drupal.org/api/drupal/core%21lib%21Drupal.php/function/Drupal%3A%3Al/8


jmolivas:
@darol100 Thanks for the PR.

PR #1508, opened 2015-12-16
Fix InvalidArgumentException on config:export:content:type

franzwilding:
This fixes the InvalidArgumentException: „The "content_type" argument
does not exist.“ I have not looked at the whole code and just tried
this out.


enzolutions:
thanks @franzwilding

Issue #1496, opened 2015-12-15
[generate:entity:content] Error in adding a Entity


NateAGeek:
The issue seems to be after creating an entity type. Adding an entity creates this error, halting the entity creation. 

Uncaught PHP Exception Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException: "The "{entity_name}" entity type did not specify a "default" form class." at docroot/core/lib/Drupal/Core/Entity/EntityTypeManager.php line 182

NateAGeek:
Related To: #1495,
Missing proper namespace and file location.

darol100:
I'm closing this issue, since is a duplicated. We can solve it at #1495

PR #1509, opened 2015-12-16
#1496 Fixed namespace forms [generate:entity:content]

mnico:


omero:
This patch doesn't work only needs a little update and remove `/Entity` in this line, https://github.com/mnico/DrupalConsole/blob/1496-fix-namespace-forms/templates/module/routing-entity-content.yml.twig#L61, to be coherent with the namespace and make the route works :smile: ... i think we only need this to merge it :+1: 
Thanks for the PR @mnico :smile:

mnico:
Thanks, I will create another pull request because I have broken this lol

PR #1512, opened 2015-12-16
[rest:enable] Fixed rest:enable to use new choice function

enzolutions:


src/Command/Rest/EnableCommand.php
```diff
@@ -68,7 +69,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $formats = $this->getSerializerFormats();
         $question = new ChoiceQuestion(
```
jmolivas: @enzolutions: We should replace this with $out->choice method

jmolivas: @enzolutions: We should replace this with $out->choice method

src/Command/Rest/EnableCommand.php
```diff
@@ -86,7 +87,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $question = new ChoiceQuestion(
```
jmolivas: same as the other comment
@enzolutions: We should replace this with $out->choice method

PR #1533, opened 2015-12-19
[generate:entity:content] Clear cache after creating a new bundle

mnico:
When I create a new bundle and then I create a new entity with this, an exception appear:
```
Drupal\Component\Plugin\Exception\PluginNotFoundException: The "entity:name_entity:new_bundle" plugin does not exist
```
This is because after creating a new bundle, the cache is not cleared.
To fix this we can use ConfigEntityBundleBase.


jmolivas:
@mnico: Thanks for the PR.

PR #1546, opened 2015-12-21
[generate:module,theme] Add https and www to d.o project homepage in composer.json

Cottser:
Minor detail but save some 301s :)


jmolivas:
@Cottser: Thanks for the PR.

Issue #1542, opened 2015-12-20
Implement DrupalStyle Class


jmolivas:
### Problem/Motivation
Standardize the output of commands.

The Symfony project released the Console Style Guide 
http://symfony.com/blog/new-in-symfony-2-8-console-style-guide

Drupal Console extends this class as `DrupalStyle`.

#### Add use keyword to Command Class
```
use Drupal\Console\Style\DrupalStyle;
```

#### Create new `DrupalStyle` instance.
Instantiating the `DrupalStyle` class as `$io`, passing the `$input` and `$output` of your command:
````
$io = new DrupalStyle($input, $output);
````

#### Replace `info` tags:

original | Replace 
------------ | ------------- 
`$output->writeln('<info>text</info>');` | `$io->info('text');` 

####  Replace `comment` tags:

| Original | Replace | 
| ------------ | ------------- |
| `$output->writeln('<comment>text</comment>');` | `$io->comment('text');` |

####  Replace `error` tags:

| Original | Replace | 
| ------------ | ------------- |
| `$output->writeln('<error>text</error>');` | `$io->error('text');` |

#### Update `success` messages:
```
$io->success('text');
```

#### Replace tables:
Original
```
use Symfony\Component\Console\Helper\Table;

$table = new Table($output);
$table->setStyle('compact');
$table->setHeaders([]);

foreach () {
  $table->addRow([]);
}

$table->render();
```
Replace:
```
$tableHeader = [];

$tableRows = [];
foreach () {
  $tableRows[] = [];
}

$io->table($tableHeader, $tableRows, 'compact');
```
> NOTE: Remove use  keyword `use Symfony\Component\Console\Helper\Table;`

### Command Namespaces status:
- [x] Config (@mnico)
- [x] Database (@jmolivas)
- [x] Develop (@jmolivas)
- [x] Exclude (@jmolivas)
- [x] **Generate** _If you are planning to work on this make sure to select one command at the time and add a comment on this issue with the command you are working on to avoid duplicated work._
  - [x] AuthenticationProviderCommand.php (@enzolutions)
  - [x] CommandCommand.php (@enzolutions)
  - [x] ConfigFormBaseCommand.php (@enzolutions)
  - [x] ControllerCommand.php (@enzolutions)
  - [x] EntityBundleCommand.php (@mnico)
  - [x] EntityCommand.php (@mnico)
  - [x] EntityConfigCommand.php (@mnico)
[...]

PR #1577, opened 2015-12-24
[yaml] #1542 Implement DrupalStyle

mnico:


src/Command/Yaml/DiffCommand.php
```diff
@@ -177,35 +170,52 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-
         // --yaml-left option
         $yaml_left = $input->getArgument('yaml-left');
         if (!$yaml_left) {
-            $yaml_left = $output->ask(
-                $this->trans('commands.yaml.diff.questions.yaml-left'),
-                null,
-                $validator_filename
-            );
+            while (true) {
```
enzolutions: Hi @mnico 

I don't understand why we need a while(true) if the ask fail the cycle never ends

src/Command/Yaml/DiffCommand.php
```diff
@@ -177,35 +170,52 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
         // --yaml-right option
         $yaml_right = $input->getArgument('yaml-right');
         if (!$yaml_right) {
-            $yaml_right = $output->ask(
-                $this->trans('commands.yaml.diff.questions.yaml-right'),
-                null,
-                $validator_filename
-            );
+            while (true) {
```
enzolutions: Same question above why a white(true)

src/Command/Yaml/MergeCommand.php
```diff
@@ -132,11 +136,18 @@ protected function interact(InputInterface $input, OutputInterface $output)
         // --yaml-destination option
         $yaml_destination = $input->getArgument('yaml-destination');
         if (!$yaml_destination) {
-            $yaml_destination = $output->ask(
-                $this->trans('commands.yaml.merge.questions.yaml-destination'),
-                '',
-                $validator_filename
-            );
+            while (true) {
```
enzolutions: same, why while(true)

src/Command/Yaml/SplitCommand.php
```diff
@@ -147,24 +152,27 @@ protected function interact(InputInterface $input, OutputInterface $output)
         // --yaml-left option
         $yaml_file = $input->getArgument('yaml-file');
         if (!$yaml_file) {
-            $yaml_file = $output->ask(
-                $this->trans('commands.yaml.diff.questions.yaml-left'),
-                '',
-                $validator_filename
-            );
+            while (true) {
```
enzolutions: while?

src/Command/Yaml/DiffCommand.php
```diff
@@ -177,35 +170,52 @@ protected function execute(InputInterface $input, OutputInterface $output)
      */
     protected function interact(InputInterface $input, OutputInterface $output)
     {
-        $validator_filename = function ($value) {
+        $io = new DrupalStyle($input, $output);
+
+        $validator_filename = function ($value) use ($io) {
             if (!strlen(trim($value))) {
-                throw new \Exception(' You must provide a valid file path.');
```
mnico: @enzolutions Hi, The use of "while" is added because I removed the throw exception in the validator

mnico: And with the "while" if the user enters an invalid path then the system will ask again for a valid path.

enzolutions:
Thanks @mnico for your contribution

PR #1581, opened 2015-12-26
[create:nodes] Add New command.

jmolivas:


jmolivas:
Command renamed to `create:nodes`
drupal-create-nodes

PR #1583, opened 2015-12-28
Hungarian for user.password.reset

rmelike:


jmolivas:
@rmelike: Thanks for the PR.

PR #1588, opened 2015-12-28
Fix the core version validation in generate:module

tstoeckler:
Currently when running `generate:module` you get a called to undefined function `8.x` because there is an incorrect parameter that is being passed to the Symfony question helper. I implemented some basic validation of the core property.

### DIFF
--- a/config/translations/en/generate.module.yml
+++ b/config/translations/en/generate.module.yml
@@ -28,4 +28,5 @@ questions:
 warnings:
     module-unavailable: 'Warning The following modules are not already available in your local environment "%s"'
 errors:
+    invalid-core: 'The core version "%s" is invalid.'
     directory-exists: 'The target directory "%s" is not empty.'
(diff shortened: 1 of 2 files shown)


jmolivas:

Since Drupal Console only supports Drupal 8 and up, this validation should only accept numbers equal and greater than 8.

dc-validate-core-version

tstoeckler:
There you go:

core-version-validate

jmolivas:
Awesome great JIT for the release

vincenzodibiaggio:
```
Enter Drupal Core version [8.x]:
 > 8
PHP Fatal error:  Call to undefined function 8.x() in /home/vincenzodb/Development/wakudoki/vendor/symfony/console/Helper/SymfonyQuestionHelper.php on line 36
PHP Stack trace:

```

I not understand, what is the correct value?

jmolivas:
`8.x` is the correct value, I will try using `8` and add comments if found anything.

tstoeckler:
@vincenzodibiaggio Please update to the latest master, that error should no longer appear

PR #1589, opened 2015-12-28
Use ContentEntityDeleteForm as a base class for content entity delete forms

tstoeckler:
Currently `ContentEntityConfirmFormBase` is being used but `ContentEntityDeleteForm` already provides the delete-specific information.

It is in fact superior to the current delete template because it properly handles entity translations.

I don't know if there is much of a use-case of overriding the delete form (i.e. I usually just reference `ContentEntityDeleteForm` in the entity annotation directly), but I went for the minimally invasive change for now.


jmolivas:

PR #1590, opened 2015-12-28
Use the language_select widget instead of a custom form element for content entities

tstoeckler:
It's preferable to use the widget so people can override this properly from the UI.

I am not sure if we want to leave in the empty `buildForm()` for people or not, but left it in for now.

Also removed the `submit()` method from the form as that is obsolete and no longer being called anyway.

And used `parent::save()` instead of calling `$this->entity->save()` manually.

### DIFF
--- a/templates/module/src/Entity/Form/entity-content.php.twig
+++ b/templates/module/src/Entity/Form/entity-content.php.twig
@@ -11,7 +11,6 @@ namespace Drupal\{{module}}\Form;
 {% block use_class %}
 use Drupal\Core\Entity\ContentEntityForm;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\Core\Language\Language;
 {% endblock %}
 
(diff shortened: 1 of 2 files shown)


jmolivas:

I did apply the changes and ran `drupal generate:entity:content` and no errors found.

But not sure how to make if this code should affect the GUI.

Do you mind to share instructions to properly test or a way to confirm, the generated code using this PR, using the UI.

tstoeckler:
Attached are screenshots of the _Manage form display_ page for an entity type generated with `generate:entity:content` before and after this commit.

manage-form-display-before

manage-form-display-after

The actual entity form looks identical. The only functional change is that before the language element was always at the bottom because it was being added at the end of the form manually after all of the widgets and now it is a widget itself. That's why I chose 10 as a default value for the weight, so that it appears last by default like it does now and can be seen in the screenshot above.

Issue #1458, opened 2015-12-12
[site:install] Recoverable fatal error.


ericmaster:
After attempting install the Standard profile through command line I'm getting the following error

`Recoverable fatal error: Argument 1 passed to Drupal\Core\Database\Driver\mysql\Install\Tasks::getFormOptions() must be of the type array, null given, called in .../core/lib/Drupal/Core/Installer/Form/SiteSettingsForm.php on line 112 and defined in .../core/lib/Drupal/Core/Database/Driver/mysql/Install/Tasks.php on line 136`

`php --version
PHP 5.6.7-1 (cli) (built: Mar 24 2015 13:43:55) 
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2015 Zend Technologies
    with Zend OPcache v7.0.4-dev, Copyright (c) 1999-2015, by Zend Technologies`

jmolivas:
I can not replicate this.

jmolivas:
DO you mind to share which, Drupal, Drupal Console and OS versions are you using?

enzolutions:
He tried Drupal 8.0.1 and Drupal Console 0.9.7 I don't remember linux version could you clarify this @ericmaster

darol100:
Closing this issue since is been commited already. Feel free to reopen if this still applys.

PR #1598, opened 2015-12-29
[site:install] Fix to enable site:install in PHP 5.6 #1458

enzolutions:


jmolivas:
Thanks @ericmaster and @enzolutions for the PR.

PR #1622, opened 2016-01-03
Use entity route providers for entities

tstoeckler:
Providing entity routes using an `EntityRouteProvider` is preferred over using a static `routing.yml` file. This changes the `generate:entity:config` and `generate:entity:content` commands to no longer add (or append to) a routing file but use entity route providers instead. Core provides (no pun intended) a default route provider which provides canonical, edit form and delete form routes automagically, so these no longer need to be specified. The generated route providers then provide add form and collection routes and for content entities the settings form route or the add page route depending on whether the entity has a bundle entity or not.

This has the following general benefits:
- Follows upstream best practice
- Reduces code duplication
- Because the definitions are dynamic they are transparent to (some) changes of the entity type definition later on, e.g. changing the entity type label.

In addition using the upstream code brings a couple of concrete improvements over the current statically generated routing YAML files:
- The canonical route has a dynamic title yielding the entity's label
- Canonical, edit form and delete form routes specify a regular expression to match the ID in case of integer IDs to improve performance

Additionally, the following bugs were fixed and improvements made that were found in testing this:
- The add form route uses an `_entity_create_access` requirement instead of a hardcoded `_permission` requirement
- The add form route falls back to the `default` form class if there is no `add` form class
- Content entities have a `collection` link template
- Link templates of content entities were fixed to point to paths under `/admin/structure`
- Config entities have an `add-form` link template
- The `collection` link template of config entities is no longer hardcoded to a specific path
- The generated access control handler for content entities triggered a notice

Also note that work is ongoing to make core's `DefaultHtmlRouteProvider` even more useful. Specifically, https://www.drupal.org/node/2600676 will add default add and collection routes so those no longer need to specified.


tstoeckler:
Just added a couple more commits.

This now also fixes the previously incorrect admin permission declaration for content entities.

bojanz:
The generated route provider always extends AdminHtmlRouteProvider, but the content entity might not be on the admin side. Worth asking a question during generation.

On a side note, we've implemented generic add-form and add-page code in the new Entity API module. The code generated here falls short. But that code won't land in core until 8.1 so I guess there's not much Console can do (without requiring Entity API along with the generated code).

tstoeckler:
I tried to mimic the current behavior as much as possible.

For example, using `AdminHtmlRouteProvider` is fairly pointless ATM because we put the admin UI under `/admin/structure` by default. You only need `AdminHtmlRouteProvider` if you want to use the admin theme for editing and the path is not under `/admin`. So I think asking for it would only make sense if we ask for the base path of the admin UI.

I didn't look at the entity API module route provider in detail, but I have no problem "stealing" the code from there. I'm not sure how involved this would be, but I guess it would be cool to check if Entity API is on for the site where the code is being set and then use that one as the base class?

In any case, would be cool to get some maintainer input ;-) I'm pretty much open in any direction

tstoeckler:
Oh, and - most importantly - thanks for the review! :-)

jmolivas:
@tstoeckler: Looks good but as @bojanz mentioned this will not work until 8.1.

I tested on `8.0.2` and got an error when trying to add a new entity using url `admin/structure/default_entity/add`:
```
Symfony\Component\Routing\Exception\RouteNotFoundException: Route 
"entity.default_entity.add_form" does not exist. in 
Drupal\Core\Routing\RouteProvider->getRouteByName() (line 191 of 
/Users/jmolivas/develop/drupal/sites/drupal8.dev/core/lib/Drupal/Core/Routing/RouteProvider.php).
```

May it worth to ask if the route is going to live under `/admin` or provide an interactive question to enter a route base path and based on that info if contains `/admin` make a decision.

tstoeckler:
Ahh, yes, the content entity needs to provide an `add-form` link in its annotation, that somehow must have gotten lost. Will fix that.

However, this pull request totally works with Drupal 8.0.* (but for the above). What @bojanz mentioned was similar to what I mentioned above that in 8.1.* the default route provider provided by core will do more automagically so we can slim the generated one down at that point. But it already [exists](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Entity!Routing!DefaultHtmlRouteProvider.php/class/DefaultHtmlRouteProvider/8) in 8.0.* So I would love to have this in Console 1.0.0 still.

jmolivas:
@tstoeckler: OK, lets have this `add-form` fixed to have it include on 1.0.0.

tstoeckler:
Just tested creating, editing and deleting both a content entity with and without bundles and for the bunlde one tried creating editing and deleting a bundle. Everything worked fine.

bojanz:
What's up with the settings form route? Seems like a weird default to have, I've developed no entity types that have a settings form so far (usually the bundle gets some).

tstoeckler:
The settings form is only for entity types which have no bundle entity type (see https://github.com/tstoeckler/DrupalConsole/blob/b55bf37ef7f7a3d961a546bb4372fc41a204a920/templates/module/src/entity-content-route-provider.php.twig#L160). Not sure if it's actually needed by default, but (like pretty much everything else) this was taken over from the current state of the routes.

bojanz:
Okay, not relevant to this PR then. Let's proceed.

jmolivas:
Yay awesome work @tstoeckler and thanks @bojanz for testing and commenting.

cweagans:
This change has made it a little more difficult to change the route paths that are set up by default. Most of it can be done by changing the annotation on the entity class, but there's still a hardcoded path in the EntitynameHtmlRouteProvider class  (see `::getAddPageRoute()`).

Couple other suggestions here:

* Generating two entities with this PR leads to a lot of duplicated code. The HTML route provider for each entity type is essentially the same. In our case, we had two separate developers use this to create two different entities in two different modules and they're largely the same code with some string changes (which is to be expected with a code generator, of course). My point is, the routing.yml that we had before was much more straightforward, but if we are *going* to use a route provider, can we at least DRY things out a bit somehow? Maybe if two Drupal Console generated entities are detected, they could use the same route provider class? Or have some kind of common api module they can depend on and only override the methods they actually need to override?

* Since so many classes are generated when creating an entity, does it make sense to also generate the corresponding unit tests? I'm going to have to write these for work anyway, so if it would be helpful, I can open a PR down the road that adds test coverage to the generated code. This could be optional, but for us, we have to maintain a certain percentage of code coverage as company policy, so having to do that by hand every time is somewhat tedious.

* Would it make sense to ask what the base routes should be for entities and bundle entities as part of the generator questions? Or can we change the defaults to more closely correspond with the Node entity? That is, my entity listing page (analogous to `/admin/content`) should live at `/admin/myentity`, and the entity form should be at `/myentity/add/bundle` (rather than `/admin/structure/whatever`).

cweagans:
@bojanz pointed out that by adding a dependency on the entity module for generated code, a lot of this boilerplate could be eliminated. I'm very in favor of that approach.

jmolivas:
Sure @cweagans

PR #1631, opened 2016-01-04
Add required PHP version to README.md

seanr:
We're getting some confusion from this so we need to flag it.  Might even want to prevent install with older PHP, but I'll leave that to smarter people.


darol100:
:+1:  For implementing this PR.

jmolivas:
:+1:  Totally agree on this one.

jmolivas:
@seanr: Thanks for the PR.

darol100:
Wooow

PR #1632, opened 2016-01-05
few more pt-br translations

lnunesbr:
=)


jmolivas:
@lnunesbr: Thanks for the PR

PR #1677, opened 2016-01-08
[console] Switch license MIT => GPL, Fix #1672

### LINKED ISSUE #1672: Determine if Drupal Console violates the GPL.
Drupal Console is MIT licensed, not sure that is allowed: https://www.drupal.org/about/licensing#q7

The question is if Drupal console is a derivative work in the same sense a Drupal module is. The safe way would be to change the license to GPLv2+.

We should probably ask the Drupal licensing group about this: https://www.drupal.org/governance/licensing-working-group

jmolivas:


darol100:
@jmolivas you are missing the License file + GPL info ?

jmolivas:
@darol100 Seems like no need to add the License file, I did check other drupal projects.

darol100:
@jmolivas , I create a PR to added the license.txt from base on this PR, adding the LICENSE.txt

PR #1680, opened 2016-01-08
[generate:entity] Fix Drupal style implementation

mnico:


mnico:
I have conflict, first i will reset my fork.

PR #1692, opened 2016-01-10
Fixes #1691

### LINKED ISSUE #1691: [generate:plugin:imageformatter]  Cache bug
### Problem / Motivation

I was running DrupalConsole from master, and I notice and when I run `generate:plugin:imageformatter` and use the image formatter with the views module. I got the following error.

> Fatal error: Class 'Drupal\bestmoduleever\Plugin\Field\FieldFormatter\Cache' not found in /Users/darrylnorris/Drupal/drupal-8.0.1/modules/custom/bestmoduleever/src/Plugin/Field/FieldFormatter/CasperImageFormatter.php on line 196

### Steps to Reproduce this bug....
- Generate a module `generate:module`
- Generate a image formatter  `generate:plugin:imageformatter`
- Create view that contain an image field.
- In the view field configuration, Change the image formatter to the module formatter. Like this...

screen shot 2016-01-10 at 12 42 44 am

### Solution
We need to add this as part of the generator - `use Drupal\Core\Cache\Cache;`

darol100:
This fixes #1691 by adding the cache namespace.

### DIFF
--- a/templates/module/src/Plugin/Field/FieldFormatter/imageformatter.php.twig
+++ b/templates/module/src/Plugin/Field/FieldFormatter/imageformatter.php.twig
@@ -19,6 +19,7 @@ use Drupal\Core\Url;
 use Drupal\Core\Utility\LinkGeneratorInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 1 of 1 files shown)


jmolivas:
@darol100: Thanks for the PR.

PR #1698, opened 2016-01-11
[translation] Japanese

ki3104:


enzolutions:
Thank you so much @ki3104

PR #1700, opened 2016-01-11
[migrate:setup] Fix to include directory files

miguel303:
Add option to load Files Directory and setup the migration


enzolutions:
Thanks @miguel303

PR #1731, opened 2016-01-14
[console] Set global option defaults, fix #1729

### LINKED ISSUE #1729: [console] Learning mode: New global option in config file
As a newbie to Drupal 8, I would like to have --learning param set for every single command, so no need to type it all the time.

jmolivas:


darol100:
Steps to reproduce this ? I uncomment the learning to true and is not running learning everytime.

screen shot 2016-01-13 at 11 34 05 pm

jmolivas:
you must run `drupal init --override` first the do the override on your home directory files

PR #1732, opened 2016-01-14
[translation] Add generate.entity.yml in hindi.

neelam-chaudhary:


enzolutions:
Thanks @neelam-chaudhary

PR #1735, opened 2016-01-14
[site:new] Add question when no site provided.

jmolivas:
This is the ouput after with this PR.
### Changes
* The `site-name` argument renamed to `directory`.
* Ask for directory if not provided as inline argument.
* Release list limited to 15.

site-new


enzolutions:
I like this changes, go and merge

PR #1749, opened 2016-01-18
Show the entity page rather than the edit form after editing an entity.

pfrenssen:
When an entity that is generated with `console generate:entity:content` is edited, the user remains on the edit form after submitting the form. This is unexpected and gives the impression to the user that the submission of the form was unsuccessful, even though a success message is shown.

The standard practice in Drupal is that a user is redirected to the full page view of the entity, let's follow this practice for the content entities we generate.

This was reported by @idimopoulos as part of a usability test for the Joinup project of the European Commission.


jmolivas:

PR #1751, opened 2016-01-18
Default to neutral for entity access

pfrenssen:
Currently the code generated for a content entity's access handler always blindly allows access for unknown operations. This is a potential security risk as it might unintentionally disclose information when a new entity operation is implemented in the future.

It would be better to default to `AccessResult::neutral()` instead of `AccessResult::allowed()`. After all, if we do not know what the operation is doing we cannot allow nor deny. This is consistent with how it is implemented in core, see for example `EntityAccessControlHandler::checkAccess()`.

This was reported by @idimopoulos as part of a code review for the Joinup project of the European Commission.


jmolivas:
@idimopoulos: Thanks for the review.

PR #1760, opened 2016-01-21
Plugin managers generated in root namespace

mglaman:
When running YAML plugin generator, the manager and manager interface are in ````\Drupal\{{ module }}```` not the Plugin namespace.

See location of https://github.com/hechoendrupal/DrupalConsole/blob/master/templates/module/src/yaml-plugin-manager-interface.php.twig


jmolivas:
@mglaman: Thanks for the PR.

PR #1763, opened 2016-01-21
[translation] Hindi

Ashwini7:
refactored the controller file translation and added translation fo entity config.


enzolutions:
Thanks @Ashwini7

PR #1777, opened 2016-01-23
correcting some EN strings and adding HU translations

balagan73:


enzolutions:
Thanks @balagan73

jmolivas:
any idea why @balagan73 @enzolutions

balagan73:
I think when merging only 2 or more files makes sense. If you try it, you get "At least two files are required" error. You can see this message in this commit, earlier it was misspelled as "At least to files are required".

PR #1784, opened 2016-01-24
Changed wording for truncating database to dropping database

balagan73:


balagan73:
The Hungarian file is under translation, I will commit it later.

PR #1787, opened 2016-01-25
Fix the argument help text for config:import.

acbramley:


jmolivas:

Issue #1739, opened 2016-01-15
[generate:entity:content] Using custom name other than DefaultEntity causes access denied


reekris:
When generating an entity with bundle support that does not use the default settings I'm getting a permission denied when trying to add a new entity of the created type.

Steps to reproduce using 0.10.5:
1.```drupal generate:module  --module="Test" --machine-name="test" --module-path="/modules/custom" --description="My Awesome Module" --core="8.x" --package="Custom" --composer```
2. ```drupal generate:entity:content  --module="test" --entity-class="Test" --entity-name="test" --label="Test" --has-bundles```
3. ```drupal module:install test```
4. Go to admin/structure/test_type
5. Create an entity type
6. Go to admin/structure/test/add

This gives an access denied error. This does not happen if the content entity was created using the default settings with --has-bundle

enzolutions:
@mnico will take care this issue

mnico:
Fixed with PR #1805

PR #1805, opened 2016-01-26
[generate:entity:content] Fixed route provider template #1739

mnico:
- This pull fix the issue #1739
- Also it set the setting form properly
- As now the [generate:module] command may not load the .module file, this command needs to detect this and then to create it if the entity has bundle.

### DIFF
--- a/templates/module/src/entity-content-route-provider.php.twig
+++ b/templates/module/src/entity-content-route-provider.php.twig
@@ -38,8 +38,8 @@ class {{ entity_class }}HtmlRouteProvider extends AdminHtmlRouteProvider {% endb
     if ($add_form_route = $this->getAddFormRoute($entity_type)) {
       $collection->add("entity.{$entity_type_id}.add_form", $add_form_route);
     }
-
 {% if bundle_entity_type %}
+
     $add_page_route = $this->getAddPageRoute($entity_type);
     $collection->add("$entity_type_id.add_page", $add_page_route);
 {% endif %}
--- a/templates/module/src/entity-content-route-provider.php.twig
+++ b/templates/module/src/entity-content-route-provider.php.twig
@@ -162,7 +162,7 @@ class {{ entity_class }}HtmlRouteProvider extends AdminHtmlRouteProvider {% endb
       $route = new Route("/admin/structure/{$entity_type->id()}/settings");
       $route
         ->setDefaults([
-          '_form' => 'Drupal\{{ module }}\Controller\{{ entity_class }}SettingsForm',
+          '_form' => 'Drupal\{{ module }}\Form\{{ entity_class }}SettingsForm',
           '_title' => "{$entity_type->getLabel()} settings",
         ])
         ->setRequirement('_permission', $entity_type->getAdminPermission())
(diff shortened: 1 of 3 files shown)


chuongle:
I fixed this by changing the MyEntity.php: `"add-form" = "/admin/structure/myentity/add"` to `"add-form" = "/admin/structure/myentity/add/{myentity_type}"`

mnico:
Thank you @chuongle, that is the right way.

templates/module/src/entity-content-route-provider.php.twig
```diff
@@ -161,7 +161,7 @@ class {{ entity_class }}HtmlRouteProvider extends AdminHtmlRouteProvider {% endb
       $route = new Route("/admin/structure/{$entity_type->id()}/settings");
       $route
         ->setDefaults([
-          '_form' => 'Drupal\{{ module }}\Controller\{{ entity_class }}SettingsForm',
+          '_form' => 'Drupal\{{ module }}\Form\{{ entity_class }}SettingsForm',
```
FatherShawn: This is good but the non-existent \Controller\ reference is in every route method.  See above ^^ Lines 140, and lines 99-100 in this display.

mnico: Hi @FatherShawn , the controller mentioned in those lines actually exists, but It is not the same that was corrected at this time.

FatherShawn: I run `generate:entity:content` and here is the list of generated files:
Generated or updated files
 Site path: /Users/shawn/Sites/entity-demo/
 1 - modules/custom/entity_test/entity_test.permissions.yml
 2 - modules/custom/entity_test/entity_test.links.menu.yml
 3 - modules/custom/entity_test/entity_test.links.task.yml
 4 - modules/custom/entity_test/entity_test.links.action.yml
 5 - modules/custom/entity_test/src/DefaultEntityInterface.php
 6 - modules/custom/entity_test/src/DefaultEntityAccessControlHandler.php
 7 - modules/custom/entity_test/src/Entity/DefaultEntity.php
 8 - modules/custom/entity_test/src/DefaultEntityHtmlRouteProvider.php
 9 - modules/custom/entity_test/src/Entity/DefaultEntityViewsData.php
 10 - modules/custom/entity_test/src/DefaultEntityListBuilder.php
 11 - modules/custom/entity_test/src/Form/DefaultEntitySettingsForm.php
 12 - modules/custom/entity_test/src/Form/DefaultEntityForm.php
 13 - modules/custom/entity_test/src/Form/DefaultEntityDeleteForm.php
 14 - modules/custom/entity_test/default_entity.page.inc
 15 - modules/custom/entity_test/templates/default_entity.html.twig

Using version 0.10.6 - I don't see entity_test/Controller/...  And if I ask phpStorm to show me all classes with DefaultEntity I get:

defaultentity-class-list

And if I open DefaultEntityHtmlRouteProvider.php and copy the class name there, `Drupal\entity_test\Controller\DefaultEntityAddController` and look for it I get no match.

Am I not looking at this correctly?

mnico: Have you created an entity without a bundle type? In this case, you are right, this command does not generate the Controller file but in the code still mentions it (specifically in RouteProvider). This not mean that it will try to use it because there are not a bundle type (see the line 91). When you create an entity with bundle type, the Controller will be created.

So I think that the Controller really should not be mentioned in the code if does not really exist. I'm going to make some changes.

FatherShawn: Ah!  That makes perfect sense - yes I was testing without a bundle based entity.  Thanks for explaining :)

jmolivas:
@mnico: Thanks for the PR.
@FatherShawn: Thanks for reviewing and commenting.

Issue #1786, opened 2016-01-25
[config:import] Should take a directory argument instead of a file.


acbramley:
It might be my naivety, but the import and export commands don't seem to line up.

We have:
* config:import - takes a file (which seems to be a gz archive, rather than a yml file)
* config:import:single - takes a config name and a file name (not sure why both are needed, they should be the same bar the extension)
* config:export - takes a directory (this works nicely)
* config:export:single - takes a config name and directory.

When we compare that to drush:
* config-export - takes a destination directory
* config-import - takes a source directory

It seems that config:export should line up with both config:import (i.e take a directory, not a file), and the drush command config-export. That would make everything consistent and mean the *:single commands are supplementary.

To get this rolling, I'll post a PR fixing up the help text for the options :)

acbramley:
https://github.com/hechoendrupal/DrupalConsole/pull/1787

jmolivas:
@acbramley: Thanks for creating the issue  
* `config:import` - The `--file` option expect a compressed file not a YAML, to import a YAML we have the `config:import:single` command.
* `config:import:single` - The `--file` option takes a single YAML file to import we have an issue to improve and support an array of files to import https://github.com/hechoendrupal/DrupalConsole/issues/1707
* `config:export ` - ;)
* config:export:single - takes a config name and directory, similar to `config:import:single` just export instead of import.

:+1: Adding a `--directory` option to `config:import` seems like must have feature, just for clarification if no `--file` option pass it will use the `CONFIG_SYNC_DIRECTORY` as default to read the config files from.

acbramley:
@jmolivas awesome, I can work on this improvement today

jmolivas:
Fixed with PR https://github.com/hechoendrupal/DrupalConsole/pull/1815

PR #1815, opened 2016-01-27
[config:import] Add directory option

acbramley:
Fixes issue #1786

Also fixes some other minor translation file issues.

I was going to take this a step further and implement the import process similar to what drush does, but I think this requires discussion and possibly a separate issue.


jmolivas:

PR #1818, opened 2016-01-27
[config:export:single] Add include dependencies option

### LINKED ISSUE #1817: [config:export:single] Should have an option for exporting dependencies as well.
### Problem/Motivation
Being able to export a single config item is useful, but it would be great to be able to export all dependencies of that config item at the same time.

### Solution
Add option for ````--include-dependencies```` which calculates the dependencies using ConfigDependencyManager ala. the Features module.

PR to follow shortly.

acbramley:
Adds the boolean option, stealing some logic from features for calculating dependencies.

Adds return types to ContainerAwareCommand and the protected class members for functions used in this command.

Fixes #1817

### DIFF
--- a/config/translations/en/config.export.single.yml
+++ b/config/translations/en/config.export.single.yml
@@ -2,6 +2,7 @@ description: 'Export single configuration as yml file.'
 arguments:
     config-name: 'Configuration name.'
 options:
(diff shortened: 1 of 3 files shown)


simesy:
Brilliant

jmolivas:
@acbramley: Thanks for the PR

PR #1822, opened 2016-01-28
[config:diff] Add new command.

### LINKED ISSUE #1821: [config:diff] Implement new command to output differences between active config and a directory.
### Problem/Motivation

It's helpful to know if anything has changed. Initial command can just show the config that has changed, could potentially extend this to actually show the difference although that could be messy.

### Solution

Implement it!

acbramley:
New command.

Outputs a table of config changes between active configuration and a directory which can either be passed in or interactively chosen from a list of core's $config_directories.

Option ````--reverse```` to get a reverse output (i.e diff SOURCE ACTIVE)

Should be extended later with a verbose option to print the actual changes, possible in another table column, more likely in a big ugly list :)

Fixes #1821


jmolivas:

Test result:
config-diff

PR #1830, opened 2016-01-29
[site:install] Account mail may be the same as defined for site mail

mnico:
This same behavior has the installer via web


jmolivas:
@mnico: Thanks for the PR.

PR #1831, opened 2016-01-29
[generate:entity:content] Remove the action add link for canonical link

mnico:
In the view of an entity, the action link "Add {entity}" is displayed. This should only be displayed in the collection.


jmolivas:
@mnico: Thanks for the PR.

PR #1855, opened 2016-02-03
[helper] detect if --uri, --target or --root options were passed and re-boot drupal.

jmolivas:


darol100:
@jmolivas what is the issue of this PR ?

novia713:
just saying, maybe a superfluous comment, but i would add a link to the book about how to set up aliases and/or suggesting a list of existing aliases.
if aliases exist something like:

```
superhero@gotham: drupal cr all
no drupal found to boot
did you mean?
[0] @drupal8
[1] @my-shiny-site
[2] @really-cool-site
```
or if aliases are not present:
```
superhero@gotham: drupal cr all
no drupal found to boot
you can find out how to set up your aliases at http://docs.drupalconsole.com/en/using/how-to-copy-configuration-files.html
```

jmolivas:
No longer applies

PR #1880, opened 2016-02-06
[translation] Catala

blasoliva:


config/translations/ca/about.yml
```diff
@@ -1,6 +1,6 @@
-description: 'Mostrar l''informació bàsica sobre el projecte Drupal Console'
+description: 'Mostrar l''informació bàsica sobre el projecte DrupalConsole'
```
enzolutions: Why do you Put Drupal Console together?

blasoliva:
Because I thought that's the correct way to do it. I'm pretty obsessed with the translation...

enzolutions:
Thanks @blasoliva

Issue #1875, opened 2016-02-05
[generate:entity:content] EntityManager is deprecated.


martinfrances107:
I love the scaffolding function....

**BUT** by using it I am creating a maintenance  problem in years to come

EntityManager is  a wrapper function MOSTLY for EntityTypeManager methods.

Many scaffold command have this problem .. to keep things small and easy to review I am going to create a new issue for each command.

darol100:
@martinfrances107 , Can you specify some eras, or which commands this is happening ?

darol100:
It seem that you already did opening multiple issues.

martinfrances107:
There are only a few template files to modify - for this issue

controller-add-page.php.twig and 
controller-entity-php.twig

I will try and sort this out over the weekend.

The console itself module makes use of EntityManager but because Drupal9 is so far off this is less of a priority.

martinfrances107:
My mistake it is important to describe what command I am altering 

it is **generate:entity:content**

martinfrances107:
I have created a pull request - which appears to solve the issue 

https://github.com/hechoendrupal/DrupalConsole/pull/1881

darol100:
It seems like we still using `getEntityManager` heavily in many commands
![screen shot 2016-04-25 at 2 35 09 am](https://cloud.githubusercontent.com/assets/6808460/14779756/73bc261a-0a8e-11e6-9deb-867de6b8e9f8.png)

martinfrances107:
Can I suggest there are two issues which I wish to separate

1) My patch deliberately modifies the twig files only  ... this for me is urgent because new code should not have known maintenance problems baked into them.

2) The maintenance hazards in the console application are separate and can factored away in there own sweet time.

jmolivas:
Totally agree with @martinfrances107 the fix was applied for the generation, the DC core code can be fixed separately.

darol100:
I will reopen https://github.com/hechoendrupal/DrupalConsole/issues/1167, which seem more for the Drupal Console core rather that the generate code.

PR #1881, opened 2016-02-06
#1875 [generate:entity:content] EntityManager is deprecated.

martinfrances107:
I have tested this by running 

drupal generate:entity:content

and using it by visiting

/admin/structure/default_entity_type/add

All seems to work fine.


enzolutions:
Thanks @martinfrances107

Issue #1515, opened 2015-12-17
Allow plug-in generation for install profiles


krlucas:
In D8 install profiles have all the power of modules so in Drupal Console they should be available as an option for generate:plugin:*.

darol100:
@krlucas , Can you provide an code example of how it should looks like ?

krlucas:
@darol100,

Sure. Here's an install profile that provides a block plug-in. 
http://cgit.drupalcode.org/sandbox-krlucas-2637020/tree/

The plug-in structure is exactly the same as it is for modules.

pfrenssen:
I need this too, we're building an install profile with custom code. I provided a fix in PR #1886.

LOBsTerr:
@jmolivas we can closed. It was fixed

jmolivas:
Closing this one. Feel free to re-open if still happening after updating to the latest version.

Issue #1717, opened 2016-01-13
Only Command classes loaded from themes


mdrummond:
Drupal Console Command classes can now be loaded in a theme. However, any supporting classes do not appear to load.

For example, a Generator class added in src/Generator is not picked up when the Command class has a use statement like so:

use Drupal\thememachinename\Generator\ClassNameForGenerator

and the Generator class has a namespace like so:

Drupal\thememachinename\Generator

It's always possible I'm setting this up wrong, but it seems like Console just isn't loading the custom Generator class. The error I'm getting is:

PHP Fatal error:  Class 'Drupal\thememachinename\Generator\ClassNameForGenerator' not found in pathtolocal/themes/thememachinename/src/Command/Generate/ClassNameForCommand.php on line 210

The error is specifically coming from drupal/src/Command/GeneratorCommand.php:31 in the stack trace.

The call stack shows this is coming from the createGenerator() method in the custom theme command. So I think the issue is that Console's Generator Command doesn't know what to do with the theme's generator class because there is no use statement for it within console. Maybe?

Although the more I experiment the more it feels like even the custom theme Command can't load the custom theme Generator.

You can find the example code in https://www.drupal.org/sandbox/mdrummond/2636658.

darol100:
:+1:  I had the same discussing with @jmolivas  yesterday about something like this. Because I try to use a Drupal plugin  (custom image formatter) in my theme and did not work. You can find more discussion about that at [Its possible to load “Image Formatter” on a theme?](https://drupal.stackexchange.com/questions/186310/its-possible-to-load-image-formatter-on-a-theme)


However, It seem like the [Bootstrap project](https://drupal.org/project/bootstrap) is doing something to autoloading some of the classes and they are using Drupal plugins in their theme.


This feature is going to be very handy specially for contrib themes, otherwise we have to have a module as dependency so we can load our classes and add logic to it that on the theme layer we cant handle.

jmolivas:
I like the idea of autoloading the commands using a better method the the implemented one. I will try to do by using composer. I will add another comment once I test this.

fmann:
I think this is an issue in modules as well as themes that try to provide custom Generators.

I'm taking a stab at #1306. No problem adding a command in layout_plugin/src/Command, but I get the same class not found error from createGenerator() as described above. I've added the Generator to layout_plugin/src/Generator.

darol100:
[...]

PR #1886, opened 2016-02-08
Support generating code for installation profiles.

pfrenssen:
Currently we can only generate code for modules. This PR allows to generate code for install profiles as well.

This fixes issue https://github.com/hechoendrupal/DrupalConsole/issues/1515.

### DIFF
--- a/src/Command/ModuleTrait.php
+++ b/src/Command/ModuleTrait.php
@@ -17,12 +17,17 @@ trait ModuleTrait
 {
     /**
      * @param \Drupal\Console\Style\DrupalStyle $io
+     * @param bool|true                         $showProfile
      * @return string
      * @throws \Exception
      */
-    public function moduleQuestion(DrupalStyle $io)
+    public function moduleQuestion(DrupalStyle $io, $showProfile = true)
     {
         $modules = $this->getSite()->getModules(false, true, true, false, true, true);
+
+        if ($showProfile) {
+            $modules[] = $this->getSite()->getProfile(false, true);
+        }
 
         if (empty($modules)) {
             throw new \Exception('No modules available, execute `generate:module` command to generate one.');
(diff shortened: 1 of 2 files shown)


darol100:
@pfrenssen , Great PR. How difficult you think will be to implement the same functionality for themes ?  We are having a similar issue with themes at - https://github.com/hechoendrupal/DrupalConsole/issues/1717

pfrenssen:
#1717 doesn't seem to be related I think?

darol100:
@pfrenssen , Even thought they have a different explanation. I think they carry the similar issue that code can only be generating for modules. Meaning that you can not generate code for themes using DrupalConsole. 

 In #1717 there is also a discussion about that some classes are not been load from the theme level, which is unrelate to this PR.

However, allowing generating code to themes is relate to this issue. Does this makes sense ?

pfrenssen:
Well this was very easy, since modules and profiles are basically the same thing in Drupal. A profile is just a module that is enabled by default and cannot be disabled, and has some additional forms and other functionality to use during the installation..

Supporting themes would be similar since they are also an "extension" in Drupal, they are handled by the `ExtensionDiscovery` class. But a theme is not the same as a module.

Since Drupal 7 it is possible to run hooks in themes, but I'm not sure if themes support everything that a module does in Drupal 8, such as event subscribers, forms, services and plugins (blocks, conditions, field widgets, ...). There are probably some differences there but I'm not very familiar with the frontend in D8 yet so I can't tell for sure.

We would need to account for the differences there are, so it would be more work. Probably we would need to look at each generator separately to see what is possible. That is out of scope for this PR though, so it can best be discussed in a dedicated issue.

jmolivas:
:+1:  on his PR.

darol100:
I have test this PR and it seem to be working fine. The only thing that I found is that only works for the profile that is install in your website. Which kinda makes sense to have it, like that.

:+1:  For merging this PR.

jmolivas:

Issue #1559, opened 2015-12-22
[database dump] Support for multiples database


darol100:
### Problem / Motivation

Database dump functionality is hardcore for MySQL. And Drupal 8 support the following databases:
- MySQL 5.5.3/MariaDB 5.5.20/Percona Server 5.5.8 or higher with InnoDB 
- PostgreSQL 9.1.2 or higher
- SQLite 3.6.8 or higher

We should add support for these database to add the ability of user backing up their database. There are other types of databases; however, they are not as popular. For now we should try to support those for database dump.

### Solution

Add support to the following database.
- [ ] MariaDB 5.5.20/Percona Server 5.5.8 or higher with InnoDB 
- [ ] PostgreSQL 9.1.2 or higher
- [ ] SQLite 3.6.8 or higher

PR #1893, opened 2016-02-09
Postgres support for database commands

ericpugh:
Added ability for sites using 'pgsql' driver to use the "database" commands. Successfully tested all "database" commands with Postgres 9.4. Restore command works on an existing DB, but throws an error if db tables don't already exist. This seems to be a general issue with Drupal 8 support for Postgres and the different method (compared to MySQL) Postgres dumps a database.

### DIFF
--- a/config/translations/en/database.log.clear.yml
+++ b/config/translations/en/database.log.clear.yml
@@ -1,4 +1,10 @@
 description: 'Remove events from DBLog table, filters are available'
+arguments:
+    event-id: 'DBLog event ID'
+options:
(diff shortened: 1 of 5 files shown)


darol100:
@ericpugh , This is relate to #1559 ?

ericpugh:
@darol100 Yes, Postgres only.

jmolivas:
:+1: on this PR.

jmolivas:
@ericpugh: Thanks for the PR.

PR #1903, opened 2016-02-10
[translation] marathi about.yml

Prabhurajn:


config/translations/mr/about.yml
```diff
@@ -1,13 +1,13 @@
-description: 'Display basic information about Drupal Console project'
+description: 'Drupal कंसोल प्रकल्पाची प्रार्थमिक माहिती प्रदर्शित करावी'
```
enzolutions: Please don't translate Drupal Console

PR #1908, opened 2016-02-11
[console] Be consistent with spacing

opdavies:


jmolivas:
@opdavies: Thanks for the PR, makes sense make it consistent and easy to read.

PR #1961, opened 2016-02-24
Fix for check db-file and db-type for sqlite options

roberto-butti:
Software version:

    Drupal Console (0.10.12) | Drupal (8.0.3)

Command executed:

    drupal site:install --langcode=it --db-type=sqlite --db-file=sites/default/files/.ht.sqlite standard

When you need sqlite and in command line you specify db-type and db-file the command in interactive mode asks you also db host.
I think that the problem is in the "if" at line number 165 of InstallCommand.php file.

    https://github.com/hechoendrupal/DrupalConsole/blob/0.10.12/src/Command/Site/InstallCommand.php#L165


jmolivas:
@roberto-butti Thanks for the PR

PR #1974, opened 2016-02-28
Fixed default values for textfields and textareas on config forms.

lucoweb:
When creating config forms, default values for fields are written on an YML file. Still needs work on field types other than textfields and textareas.


jmolivas:
@lucoweb Thanks for the PR

PR #1976, opened 2016-02-29
Fix all coding standard issues and some typos in views field plugin generation.

eiriksm:
Hey there.

First of all, thanks so much for this tool. It's pretty much the most awesome ever.

Just noticed a non-ideal output from the generation of a views field plugin. Here is a PR that makes it output code that passes Drupal coding standards and fixes one typo (generte => generate).

I see in the travis file that there is some effort to try and check Drupal CS on build. Will look at that, and see if I can help with something there :)

Have a great day!


jmolivas:
@eiriksm Thanks for the PR.

PR #1977, opened 2016-02-29
Make example chain create the wanted states in REST resource.

eiriksm:
The config provided in the chain example does not create the intended REST states. 

Here is a PR fixing that.


eiriksm:
Hm, the test fail actually highlights the problem I guess :)

I can rebase this PR if we get the actual template fix merged first, then this should also pass.

PR #1978, opened 2016-02-29
Add missing label to config and content entities in the example chain.

eiriksm:


jmolivas:
@eiriksm Thanks for the PR

PR #1980, opened 2016-02-29
Fix coding standards for controller generation.

eiriksm:


jmolivas:
@eiriksm Thanks for the PR.

PR #1981, opened 2016-02-29
Fix coding standards for REST resource generation.

eiriksm:
OK, here I have actually changed a couple of things:

- There were to unused 'use' statements. `AccessDeniedHttpException` and `HttpException`.
- I decided to just use the former, since having it in a comment would not look very good while still passing standards for inline comments. Please chime in, if you disagree.
- I decided to just remove the latter, since I (personally) don't see what it adds of value to the user. Since we already show how to throw an AccessDeniedHttpException, I think people would figure these things out?
- Decided to get the logger service namespaced with our custom module (example). This does not relate to coding standards, so if you want me to add another PR for that line (or if you disagree), I am fine with that.

Thoughts?


jmolivas:
@eiriksm Thanks for the PR.

PR #1982, opened 2016-02-29
Fix coding standards for config entities.

eiriksm:
OK, this change looks a little uglier actually. But it produces better output code, which I think is more important.

If we have the endblock on the next line, it will produce 2 blank lines, and the sniffer will complain.


jmolivas:
@eiriksm Thanks for the PR nice catch. 

Coding style is more important the making templates look prettier.

PR #1983, opened 2016-03-02
Fixes class_name in rest.php.twig @file block.

steveoliver:
The trailing {{ module_name }} should be {{ class_name }} in rest.php.twig.


jmolivas:
@steveoliver Thanks for the PR, nice catch.

Issue #1968, opened 2016-02-26
[state:delete] Remove state


darol100:
### Problem / Motivation
We need to add the ability of deleting/removing state from a Drupal website.

### Solution
- Add the ability of removing states.

darol100:
This have been solve. Closing it.

PR #1985, opened 2016-03-02
[state:delete] Add new command, working progress of #1968

darol100:
This command works except the error if there is not that state is not available in the system.

I have comment this code out ....

``````
                $keyValue = $this->getService('keyvalue');
                if (!$keyValue->exists($name)) {
                    $io->error(
                        sprintf(
                            $this->trans('commands.state.delete.messages.state-not-exists'),
                            $name
                        )
                    );

                    return 1;
                }

``````


darol100:
Ready for testing.

PR #1993, opened 2016-03-06
Update document Readme

isimgt:
Update and reorganize the readme document, I think that the old document it was and now is more clean.


jmolivas:
@isimgt: Thanks for the PR.

PR #1994, opened 2016-03-06
Fixing configuration name

darol100:
In the `config:import:single` did not ask a question about which configuration name. There was some minor bugs. In addition, I have rename the config-name to name.


jmolivas:
@darol100 can you please sync this with latest changes on master?

PR #1997, opened 2016-03-06
[helper] Autoload Generator classes defined within themes.

mdrummond:
# Problem/Motivation
This addresses https://github.com/hechoendrupal/DrupalConsole/issues/1717 so that Commands added within themes can make use of Generators added within Themes.

# How to Reproduce

Within a custom theme for Drupal 8, add a Command class that makes use of a Generator class. https://www.drupal.org/sandbox/mdrummond/2636658 is a good example. Right now, the Generator class is defined in the same file as the Command class. Remove the Generator definition from the Command class and uncomment the use statement for the Generator class that is defined in a separate file.

# Solution
Within the Command Discovery process, this looks for Generator classes within themes and includes those classes.

Worked on this with Jesus at Drupal Camp Florida.

# Minimum Requirements
This requires Drupal 8.


jmolivas:
@mdrummond: Thanks for the PR.

PR #1999, opened 2016-03-07
[generate:plugin:ckeditorbutton] New command

dinarcon:
Please review.

### DIFF
--- a/Test/DataProvider/PluginCKEditorButtonDataProviderTrait.php
+++ b/Test/DataProvider/PluginCKEditorButtonDataProviderTrait.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Drupal\Console\Test\DataProvider;
+
+/**
+ * Class PluginCKEditorButtonDataProviderTrait
+ * @package Drupal\Console\Test\DataProvider
+ */
+trait PluginCKEditorButtonDataProviderTrait
+{
+    /**
+     * @return array
+     */
+    public function commandData()
+    {
+        $this->setUpTemporaryDirectory();
+
+        return [
+          ['Foo',  'foo' . rand(), 'foo', 'bar'],
+        ];
+    }
+}
--- a/Test/DataProvider/PluginCKEditorButtonDataProviderTrait.php
+++ b/Test/DataProvider/PluginCKEditorButtonDataProviderTrait.php
@@ -16,7 +16,7 @@ trait PluginCKEditorButtonDataProviderTrait
         $this->setUpTemporaryDirectory();
 
         return [
-          ['Foo',  'foo' . rand(), 'foo', 'bar'],
+          ['Foo',  'foo' . rand(), 'foo', 'bar', 'Baz'],
         ];
     }
 }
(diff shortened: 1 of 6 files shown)


darol100:
@dinarcon , This is a great start. I'm able to drag and drop the of the icon at the text format configuration(`admin/config/content/formats/manage/full_html`).  However, I'm not able to see it at the editor when I'm creating a page. And btw I'm selecting the full_html text format. 

I have try it out with the Youtube add on as well as with the `loremipsum` add on.  On Drupal 8.3 and 8.5.

I have create a sandbox in Drupal.org - https://www.drupal.org/sandbox/darol100/2683885 so you can checkout if I'm doing something wrong or missing something something.

Here is the addon page - http://ckeditor.com/addon/loremipsum

dinarcon:
Hi @darol100, thanks for your review! Your test revealed an oversight which has been fixed already. A new option is required for the command to work. I have provided better instructions to set them correctly. In particular:

- The plugin ID corresponds to the CKEditor plugin name. It is the first argument of the CKEDITOR.plugins.add() function in the plugin.js file.

- The button name corresponds to the CKEditor button name. They are the first argument of the editor.ui.addButton() or editor.ui.addRichCombo() functions in the plugin.js file.

In the case of the loremipsum ckeditor button you are working on, the following command will generate the appropriate code:

`drupal generate:plugin:ckeditorbutton --module=ckeditor_loremipsum --class=LoremipsumCKEditorButton --label='Lorem ipsum' --plugin-id=loremipsum --button-name=Loremipsum`

Please note that following CKEditor's recommendations, the plugin code is suggested to be placed in a 'js/plugins' folder. And make sure to update the path to your icon accordingly. This varies from among plugins.

Please test again.

darol100:
> Please note that following CKEditor's recommendations, the plugin code is suggested to be placed in a 'js/plugins' folder. And make sure to update the path to your icon accordingly. This varies from among plugins.

Even thought CKeditor recommendation is inside of the module, Drupal recommendation is to have it on central location just in case some other module was to use it. In D7 we use to store our external libraries at `/sites/all/libraries` in D8 that location has changed to `/libraries` So, I think we should store the location at `/libraries` to follow Drupal best practices.

@dinarcon What do you think ?

dinarcon:
@darol100 It is possible to change the code to point to the location of preference. The default is just a suggestion which is based on both CKEditor's recommendations and Drupal's own plugin implementations. Contrib is also following this pattern. Maybe we can open an issue on drupal.org to further discuss this.

In terms of functionality, is the current implementation showing the button on the editing screen and working as expected? It would be great to test other CKEditor plugins to verify this works for all cases.

darol100:
> In terms of functionality, is the current implementation showing the button on the editing screen and working as expected?

Yes, for the lorempisum library seem to be working fine. I will try it out other libraries to fully test this PR.

dinarcon:
@darol100 great, thanks for your review!

jmolivas:
@dinarcon Thanks for the PR.

jmolivas:
@darol100 Thanks for reviewing this.

PR #2001, opened 2016-03-07
Adding interaction into the state override command.

darol100:
Adding the interaction function to the state command for consistency across other commands.


jmolivas:
@darol100 Thanks for the PR.

PR #2004, opened 2016-03-09
Project Clean up...

daceves:
* Removed print_r()
* Removed exit(). Avoid using exit, throwing exception instead.
* Changing logical operators from 'and' to '&&' to avoid potential unintended behavior
* Adding 'use' for Symfony\Component\Yaml\Exception\ParseException; To avoid confusion with Symfony\Component\CssSelector\Exception\ParseException
* Removed executable permission on config/translation/vn/*.yml files


jmolivas:
@daceves: thanks for the PR, I will be looking at this today

darol100:
:+1:  For this PR.  I do not see anything wrong with it.

jmolivas:
Actually is a pretty nice PR ;)

daceves:
Thanks. :+1:

src/Command/Self/UpdateCommand.php
```diff
@@ -124,6 +124,6 @@ protected function execute(InputInterface $input, OutputInterface $output)
         // can't read files properly from the overwritten Phar, or perhaps it's
         // because the autoloader's name has changed. We avoid the problem by
         // terminating now.
-        exit;
+        throw new Exception('Unexpected instantiation');
```
jmolivas: If we do not have an exit here the self-update command output an exception after updating the project.

jmolivas:
I sent another PR https://github.com/hechoendrupal/DrupalConsole/pull/2033 including your changes + other improvements.

PR #2014, opened 2016-03-10
[helper] replace entityManager to entityTypeManager

gnuget:
**Problem/Motivation**

EntityManager is deprecated and will be removed  before drupal 9.0.x 

In this PR  I just replace entityManager  in favor of  entityTypeManager.

We could even replace: 

`$node = $this->entityTypeManager->getStorage('node')->create(...)`

to
`$node = Node::create(...)`

I will create a another PR later.

### DIFF
--- a/src/Helper/DrupalApiHelper.php
+++ b/src/Helper/DrupalApiHelper.php
@@ -8,7 +8,6 @@
 namespace Drupal\Console\Helper;
 
 use Symfony\Component\DomCrawler\Crawler;
-use Drupal\Console\Helper\Helper;
 use Drupal\Console\Utils\Create\Nodes;
 use Drupal\Console\Utils\Create\Terms;
 use Drupal\Console\Utils\Create\Vocabularies;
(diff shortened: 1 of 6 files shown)


jmolivas:
@gnuget: Thanks for the PR.

I like injecting the object better thank using static methods.

And looking at the code for ` Node::create(...)` is using the deprecated method `entityManager`.
```
  /**
   * {@inheritdoc}
   */
  public static function create(array $values = array()) {
    $entity_manager = \Drupal::entityManager();
    return $entity_manager->getStorage($entity_manager->getEntityTypeFromClass(get_called_class()))->create($values);
  }
```

gnuget:
Well using the  static method is up to you.

In my opinion the static method  is clearer about what it do, it is part of the Drupal API,  it is shorter,  and eventually that method will be updated to use entityTypeManager instead of entityManager so you won't need to be worried about update it later (https://www.drupal.org/node/2674678)

But even if  you decide to not use the static method, I would really recommend to try to use only non-deprecated  methods/classes even if the core still use them, because some people will eventually use the  console  code as base to do their own code and in d.o there are already some issues created to update the core itself. 

BTW just recently started using the Console, and I'm sure to many people already told you this but awesome work!!!   😃

jmolivas:
I tested this and looks good.

PR #2024, opened 2016-03-15
minor grammatical update

justinlevi:
Changing `...It defines with what function...` to read as: 

`..It defines what function...`


jmolivas:
@justinlevi Thanks for the PR.

PR #2038, opened 2016-03-18
[console] Remove extra slashes in the autoloader include

opdavies:


jmolivas:
@opdavies Thanks for the PR.

PR #2039, opened 2016-03-18
[helper] Removed HelperTrait import

opdavies:
This is not needed as both files are within the same namespace.


jmolivas:
@opdavies Nice catch and thanks for the PR.

PR #2042, opened 2016-03-18
Adding state delete back.

darol100:


jmolivas:
@darol100 Thanks for the PR

PR #2043, opened 2016-03-18
[site:mode] Invalid message on copy problem

Peekmo:
Hello,

This PR corrects typo on a path when there's a copy problem on services.yml (I though the command wasn't working because of that, but it was just the message that is wrong).

Regards,
Peekmo


jmolivas:
@Peekmo Thanks for the PR.

Issue #2057, opened 2016-03-23
Create a FieldTrait to gather a list of fields for commands


cosmicdreams:
Last night on Drupal and Coffee we started making a FieldTrait that would gather up all the fields from a Drupal site so that we could have a wizard pick one and then pass that info to some kind of field generator.  

Kind of like FormTrait and the Form Generator.  

I think it would be really impressive if we could leverage a FieldTrait to create a command where you generate a content type / custom entity / thingy and then generate all the fields that should be associated with it.  

A common frustration during the content modeling part of a project is having to click through all the pages for fields.  If we had a command that did this then we'd be able to make all the decisions for the fields up front and then batch all the config generation.

PR #2058, opened 2016-03-23
Add a FieldTrait

cosmicdreams:
Obviously this needs improvement, but I wanted to share the code we worked up last night for FieldTrait.  This is just a copy of FormTrait and then we tried to make it look for a list of fields instead of a list of form api definitions.  We didn't finish because we need to debug this to find the right way to get that list of fields.

The hurdles ahead:
1. figure out how to get a list of fields
2. Maybe we need to have a factory that can pick the right generator for field config
3. Maybe we need to alter the wizard's workflow to provide a "Shortcut" were it just accepts all the defaults.

### DIFF
--- a/src/Command/FieldTrait.php
+++ b/src/Command/FieldTrait.php
@@ -1,0 +1,176 @@
+<?php
+
+/**
+ * @file
+ * Contains Drupal\Console\Command\FormTrait.
+ */
+
+namespace Drupal\sammy\Command;
+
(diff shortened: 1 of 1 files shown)


jmolivas:
@cosmicdreams :+1: on this feature I will take a look and try to keep improving it.

cosmicdreams:
This is related to #2057

jmolivas:
@cosmicdreams  Any plans to keep working on this?

Issue #2069, opened 2016-03-27
Allow "choiceNoList" to set custom value


darol100:
### Problem / Motivation
**choiceNoList** does not allow you to add parameters that are not already defined. We are using **choiceNoList** inside of config:import and does not allow user to enter custom configuration and it should.

### Solution
- Add the ability of add custom values on **choiceNoList** .

jmolivas:
We have the option on choiceNoList to accept a null Value, we need to improve the helper to accept any arbitrary value as well.

PR #2070, opened 2016-03-27
Improvement for config:import:single. 

### LINKED ISSUE #2054: [config:import:single] Doesn't work
I tried different combinations of params for `drupal config:import:single` command but in all cases it just hangs forever. I also passed `-v` but this didn't produce any output either.

darol100:
Fixes #2054.

This also added the ability of asking question for the path file.

### DIFF
--- a/config/translations/en/config.import.single.yml
+++ b/config/translations/en/config.import.single.yml
@@ -1,6 +1,10 @@
 description: 'Import the selected configuration.'
 arguments:
-    name: 'Configuration name.'
-    input-file: 'Path to the import files.'
+    name: 'Configuration name'
+    file: 'Path to the import file'
+questions:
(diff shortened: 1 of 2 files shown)


darol100:
Here is an example of the command.

config-import-single-improvements


There are 2 things to consider now. We need to fix https://github.com/hechoendrupal/DrupalConsole/issues/2069 on the same release otherwise we are going to create a huge bug.

And another thing to consider is ....

https://github.com/hechoendrupal/DrupalConsole/issues/2054#issuecomment-202068231
> The file path should not be mandatory, by default the file can be picked from sync/ as this is done by config:export:single command.

@jmolivas can you confirm if this is actually true ? As far I know, this only happen on `config:import`.

PR #2072, opened 2016-03-27
Fixes #2071

### LINKED ISSUE #2071: [translation] Use double quotes when %s is used it.
### Problem / Motivation

There is been some inconsistency between commands, where some of them use `%s` and some of them use `"%s"` at the translation level.
### Solution

Rename `%s` to `"%s"`.

darol100:
Use double quote. See more at #2071


darol100:
@jmolivas another hacking solution for this is fixed in English and run the script to make re-write those lines in other languages ?

PR #2073, opened 2016-03-27
Module uninstall force

### LINKED ISSUE #1745: [module:uninstall] add force uninstall option
## Motivation
During intensive development we sometimes need to re-install modules that are dependent upon by other modules.

## Solution
Adding  an option to the module uninstall command to ignore dependencies

darol100:
fixes #1745.


darol100:
forceuninstall

Issue #2061, opened 2016-03-23
Avoid using symfony/dependency-injection


greg-1-anderson:
Drupal core uses different versions of symfony/dependency-injection in the 8.1.x line than it does in the 8.0.x line. Even though these two versions are very close, and API-compatible, they definitely do NOT work well together.

The upshot of this is that if DrupalConsole uses symfony/dependency-injection for its DI container, then it will not be possible for a global install of it to work with both a Drupal 8.0.x version and a Drupal 8.1.x version, because the version of symfony/dependency-injection used in DrupalConsole can be compatible with at most one of these. This situation might get worse in the future; it appears that the version of Symfony never changed, even by a bugfix release, in the Drupal 8.0.x line, but I don't know that there is a policy on this. If there isn't, then it is possible that there might be a bugfix release in Drupal 8.1.x that bumps up the version of symfony/dependency-injection, making things even more difficult.

Of course, overall, site-local installs of Drush and DrupalConsole are the best way to go, but there are still a lot of folks who use global installs of these tools, so we're trying to keep things as compatible as possible. To make things more flexible for Drush, we selected [league/container](https://github.com/thephpleague/container) for our DI needs.  After using it for a little while, I find that I prefer it to symfony/di. Converting wouldn't be too hard; you'd just need to convert your compiler passes to inflectors -- exact techniques might vary depending on what you're doing.

There is more discussion at https://github.com/drush-ops/drush/pull/2018, especially near the end.

joelpittet:
Cross posting with a core issue that relates:
https://www.drupal.org/node/2690835

greg-1-anderson:
Related: https://github.com/drush-ops/drush/pull/2103

greg-1-anderson:
Now that Drupal 8.1 has a rc release, I was wondering what DrupalConsole's strategy was moving forward. Will any effort be made to allow a single DrupalConsole executable to work with both 8.0.x and 8.1.x releases of Drupal?

greg-1-anderson:
Related:  #2022

citlacom:
Is there a fix for this? Should I upgrade Drupal Console to resolve this problem?

greg-1-anderson:
[...]

PR #2081, opened 2016-03-31
Register annotated commands in Drupal Console.

greg-1-anderson:
This PR adds annotation command support to DrupalConsole.

Any command file `FooCommand.php` may now contain either a DrupalConsole Symfony Console Command implementation, in which case it works as it always has, or, alternately, if the class FooCommand does not extend Drupal\Console\Command\Command, then it is presumed that the instance is an annotation command, and all of its public methods are registered as DrupalConsole commands.

The purpose of this is to allow Drupal modules to implement simple, loosely-coupled commandline tool interfaces. These commands might be implemented by extending [Robo](https://github.com/Codegyre/Robo), although this is not required.

I was planning on converting the DrupalConsole InitCommand into an annotation class, simply to demonstrate how that is done; however, since this command is already tightly coupled with Drupal\Console\Command\Command helpers, `getModule()`, etc., doing this would be counterproductive; in general, it is a non-goal to convert core DrupalConsole commands into annotation commands. For a typical simple Drupal module, though, whose main purpose is to provide a CLI interface that simply calls a method or two in the module, the annotation class will be a win.

Follow development in Drush at https://github.com/drush-ops/drush/issues/2103.

DrupalConsole checklist (to consider / discuss):

- [x] Allow Command object to be passed in as a parameter to task methods that need it. https://github.com/consolidation-org/annotation-command/pull/4
- [x] Allow $input and $output objects to be passed in as parameters. https://github.com/consolidation-org/annotation-command/pull/5
- [ ] Use league/container so that we have use of inflectors https://github.com/hechoendrupal/DrupalConsole/issues/2061
- [ ] Use a styled PSR-3 logger https://github.com/consolidation-org/log
- [ ] Consider including Robo in composer.json, so Annotated commands have access to tasks.
- [ ] Translation strategy
  - [ ] Annotation strings
  - [ ] PSR-3 logger strings

### DIFF
--- a/src/Command/Command.php
+++ b/src/Command/Command.php
@@ -2,7 +2,7 @@
 
 namespace Drupal\Console\Command;
 
-use Symfony\Component\Console\Command\Command as BaseCommand;
+use Consolidation\AnnotatedCommand\AnnotatedCommand;
 use Symfony\Component\Console\Helper\HelperSet;
 use Drupal\Console\Helper\HelperTrait;
 
@@ -10,7 +10,7 @@ use Drupal\Console\Helper\HelperTrait;
  * Class Command
  * @package Drupal\Console\Command
  */
-abstract class Command extends BaseCommand
+abstract class Command extends AnnotatedCommand
 {
     use HelperTrait;
 
--- a/src/Command/MemeCommand.php
+++ b/src/Command/MemeCommand.php
@@ -1,6 +1,10 @@
 <?php
 namespace Drupal\Console\Command;
 
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Drupal\Console\Command\Command;
+
 /**
  * Ported by Greg Anderson.
  * Original code Copyright (C) 2013 by Muazzam Ali
@@ -23,7 +27,7 @@ namespace Drupal\Console\Command;
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-class MemeCommands
+class MemeCommand extends Command
 {
     private $upperText;
     private $lowerText;
@@ -35,19 +39,30 @@ class MemeCommands
     /**
      * Generate a meme
      *
+     * @command generate:meme
      * @param string $topMessage The message to place at the top of the image.
      * @param string $bottomMessage The message to place at the bottom of the image.
      * @param string $image The path to the meme image to use.
+     * @default $image meme.png
      * @option string $out The file to write the generated meme to.
+     * @default $out generated-meme.png
      * @usage generate:meme "I don't always make memes" "But when I do, I use the command line" my-source-image.png
      */
+    public function execute(InputInterface $input, OutputInterface $output)
+    {
+        $topMessage = $input->getArgument('topMessage');
+        $bottomMessage = $input->getArgument('bottomMessage');
+        $memeImage = $input->getArgument('image');
+        $outputFile = $input->getOption('out');
+
+        return $this->generateMeme($topMessage, $bottomMessage, $memeImage, $outputFile);
+    }
+
     public function generateMeme(
         $topMessage,
         $bottomMessage,
         $memeImage = 'meme.png',
-        $options = [
-            'out' => 'generated-meme.png',
-        ]
+        $outputFile = 'generated-meme.png'
     ) {
         $finfo = new \finfo(FILEINFO_MIME);
         if ($memeImage[0] != '/') {
@@ -67,7 +82,9 @@ class MemeCommands
 
         $this->setUpperText($topMessage);
         $this->setLowerText($bottomMessage);
-        $this->processImg($options['out']);
+        $this->processImg($outputFile);
+
+        return 0;
     }
 
     protected function locateFont($fontNames)
--- a/src/Command/Settings/InitCommand.php
+++ b/src/Command/Settings/InitCommand.php
@@ -19,24 +19,13 @@ use Drupal\Console\Style\DrupalStyle;
 class InitCommand extends Command
 {
     /**
-     * {@inheritdoc}
-     */
-    protected function configure()
-    {
-        $this
-            ->setName('settings:init')
-            ->setDescription($this->trans('commands.settings.init.description'))
-            ->addOption(
-                'override',
-                null,
-                InputOption::VALUE_NONE,
-                $this->trans('commands.settings.init.options.override')
-            )
-            ->setAliases(['init']);
-    }
-
-    /**
-     * {@inheritdoc}
+     * Init command
+     *
+     * Copy configuration files to user home directory.
+     *
+     * @command settings:init
+     * @option boolean $override Override configurations files
+     * @aliases init
      */
     protected function execute(InputInterface $input, OutputInterface $output)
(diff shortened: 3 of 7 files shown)


greg-1-anderson:
Replaced old technique with something better. No longer overloads the "*Command.php" namespace, and instead looks for command files following their own rules.

Untested, but on the right track.

greg-1-anderson:
The latest commit (12daaa6) works great with the [sample cli command I wrote (ported) for the default_content module](https://www.drupal.org/node/2703131). Search paths and naming conventions still need to be discussed and finalized. I am not set on (or even very happy with) the placeholder "CliTools" folder name that I am using now; just picked something quickly to get started.

Does not work with Drush yet, but it will only be a small effort to fix up the PR there as well; I'm not going to do that tonight, but will have it working in short order.

greg-1-anderson:
Now also works with Drush.

greg-1-anderson:
Discussion on search paths and naming conventions for command files at https://github.com/consolidation-org/annotation-command/issues/12

greg-1-anderson:
Just for fun, I added a generate:meme example command. We can take this out prior to merge if desired; in the interim, it is useful to have a test command bundled with the PR.

Uses Impact font on a Mac. If running on some other platform, download Anton.ttf from https://googlefontdirectory.googlecode.com/hg/ofl/anton/ and place Anton.ttf in ~/.fonts.

jmolivas:
@greg-1-anderson  Nice usage of annotations, we have been discussing using them before as you can see on issue [Use annotations in Commands](https://github.com/hechoendrupal/DrupalConsole/issues/478) and the related PR [Annotations in Drupal Commands](https://github.com/hechoendrupal/DrupalConsole/pull/479).

Our proposal for using annotation was for other purpose. We wanted annotations to define any behavior not related to Symfony/Console Command and make our commands as much as close to the official standard. [Creating a basic Command - Symfony documentation](http://symfony.com/doc/current/components/console/introduction.html#creating-a-basic-command)

#### As stated on issue definition
> The Drupal commands interact with more stuffs that the commands in Symfony, in Drupal, one command can be have many dependencies with other modules or themes or any other scope in your installation.

> This approach is to use Annotations to provide metadata on each command to specify dependencies or check another variables in Drupal or any other stuff that interacts with the console.

> For example:
```php
/**
 * @DrupalCommand (
 *      dependencies = {
 *          "simpletest"
 *      },
 * )
 */
class TestDebugCommand extends ContainerAwareCommand { ... }
```

With that said, after discussing with the other project maintainers we don’t agree with creating a new Drupalisim or a new Drupal way to write Symfony/Console based Commands when there is a well accepted and defined standard provided and supported by the community.

We really like the idea about decouple code create components to reuse between different Drupal CLI tools.

greg-1-anderson:
Well, this is very disappointing, and more than a little ironic; however, I think there is nothing to do here at this point but close the PR and move on.

greg-1-anderson:
I have an idea on how we could configure Symfony/Console commands without diverging from the Symfony API.  Re-opening so I can push another commit here to demonstrate.

src/Command/MemeCommand.php
```diff
@@ -0,0 +1,308 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+class MemeCommand extends Command
```
greg-1-anderson: Commands extend Drupal\Console\Command\Command, just as before. These are Symfony commands.

src/Command/MemeCommand.php
```diff
@@ -0,0 +1,308 @@
...
+     * @command generate:meme
+     * @param string $topMessage The message to place at the top of the image.
+     * @param string $bottomMessage The message to place at the bottom of the image.
+     * @param string $image The path to the meme image to use.
+     * @default $image meme.png
+     * @option string $out The file to write the generated meme to.
+     * @default $out generated-meme.png
+     * @usage generate:meme "I don't always make memes" "But when I do, I use the command line" my-source-image.png
+     */
+    public function execute(InputInterface $input, OutputInterface $output)
```
greg-1-anderson: The configure() method is replaced with annotations on the execute() method. These annotations could also go on the class docblock; I'm not sure which is better.

src/Command/Settings/InitCommand.php
```diff
@@ -19,24 +19,13 @@
...
-     * {@inheritdoc}
+     * Init command
+     *
+     * Copy configuration files to user home directory.
+     *
+     * @command settings:init
+     * @option boolean $override Override configurations files
+     * @aliases init
      */
     protected function execute(InputInterface $input, OutputInterface $output)
```
greg-1-anderson: Again, porting existing Drupal Console commands to use annotations is not required; however, I did convert the settings:init command simply to allow configuration via annotations to be compared side-by-side with configuration via php method calls.

src/Command/Command.php
```diff
@@ -2,15 +2,15 @@
...
+use Consolidation\AnnotatedCommand\AnnotatedCommand;
 use Symfony\Component\Console\Helper\HelperSet;
 use Drupal\Console\Helper\HelperTrait;
 
 /**
  * Class Command
  * @package Drupal\Console\Command
  */
-abstract class Command extends BaseCommand
+abstract class Command extends AnnotatedCommand
```
greg-1-anderson: Drupal\Console\Command now extends from AnnotatedCommand.  AnnotatedCommand extends Symfony\Component\Console\Command\Command, and will behave exactly like a Symfony/Console Command unless there are annotations present that configure the command.  Even when annotations are used, the base class will still operate the same way, calling `execute` with `$input` and `$output` variables when the command is used.

greg-1-anderson:
As you can see from the diffs in this PR, the code now conforms to the Symfony/Console command API exactly, with annotations being used only as an optional configuration method. Having this in place also gets us one step closer to being able to use annotations to specify dependencies, as referenced in the comments above.

The importance of this PR extends beyond the simple application of annotations for configuration. Using a common base class from Consolidation-org will allow us to do things such as share the [output formatters](https://github.com/consolidation-org/output-formatters) library between Drush and Drupal Console. Drupal Console has a similar feature, but the consolidation output formatters are more complete, and using a common implementation will allow us to present a uniform interface for Drupal modules to use.

I hope that you will find this new implementation more palatable than the previous one. Using annotations in Drupal Console core is not required; accepting this PR would allow module developers to use or not use them as they saw fit.

jmolivas:
Looking at this, thanks for the update @greg-1-anderson

jmolivas:
Support for annotations merged on PR https://github.com/hechoendrupal/DrupalConsole/pull/2642

PR #2103, opened 2016-04-06
[config:override] Adding interaction

### LINKED ISSUE #2091: [config:override] Interact and Question
### Problem

We should provide a question when the module is going to be the user is going to override a configuration. This should ask any configuration is available in the system and the key that is going.
### Solution

We need to ask 3 question:
- Name of the config with interaction from the system.
- Name of the key value with an interaction from the config selected.
- Ask for the name of new value

darol100:
Fixes #2091


darol100:
config-override-interact


However, if we pass a config that is not in the system is going to throw this error. 

> Recoverable fatal error: Argument 2 passed to Drupal\Console\Style\DrupalStyle::choiceNoList() must be of the type array, null given, called in /Users/darrylnorris/Desktop/DrupalConsole/src/Command/Config/OverrideCommand.php on line 58 and defined in Drupal\Console\Style\DrupalStyle->choiceNoList() (line 38 of /Users/darrylnorris/Desktop/DrupalConsole/src/Style/DrupalStyle.php).

screen shot 2016-04-05 at 7 41 28 pm

There are two options to solve this issue:
 - We could not allow overriding any config that is not in the system. And this will throw an error if the config is not in the system.
 - Or we can make behave the same way as `state:override`, which ends up setting that as new config with a key and value.

darol100:
Closing this PR, I will reopen another one with clean commits

PR #2104, opened 2016-04-06
[module:uninstall] Interact and Question 

### LINKED ISSUE #2074: [module:uninstall] Interact and Question
### Problem / Motivation
We should provide a question when the module is going to be uninstalled. This should ask which module is going to be uninstalled and it should provide a list of modules that is enabled.

### Solution
- Provide the interact function with the questions.

darol100:
Fixes #2074
module-uninstall-interact

A nice improvement that we could do is adding the ability of multiples modules with commas. I personally do not like the behavior from the `module:install` where is going to loop and keep asking for the same question over and over again.


darol100:
Closing this PR, I will reopen another one with clean commits

PR #2109, opened 2016-04-07
Novia713 generate:doc:cheatsheet

novia713:
Right now i think as first aproximation as code contributor this PR is enough.
It's is working, it generates a PDF. The only fault i see is that it's not using Twig.
I create the HTML (wkhtmltopdf generates the PDF over a HTML file) programatically. I think other people can contribute doing a twig file and rendering it ...
¿Did you see the code? ¿What do you think? :smile: 

The only mandatory argument is «path» in order the command to know where to generate the PDF.
Once twig were integrated in the future, a --format argument can be given, for getting a HTML output as option.
There is another argument --wkhtmltopdf to say the command where the wkhtmltopdf binary lives. If not given, "/usr/bin/wkhtmltopdf" is guessed.

I think a new key as _description-for-cheatsheet_ for every command is not necessary, I'm printing right now the command description and it's fine.

Further in time, perhaps we can do a .yml config file for selecting which commands to print and which not. Even, giving the user this choice. This should be easy, given I'm iterating already a list of arbitrary commands (all except yaml & translation).

Please keep an eye on the code, I'm expecting your considerations :smile: 

--

In other terms, I'm really interested in doing tests. Many classes are still without them, for example, _PHPUnit_ says _GeneratorPluginBlockCommandTest_ has tests not implemented yet. 
I would do it, but i don't know which file is missing .....
So a little explanation on the testing process would be a great help for me.

Regards.


jmolivas:
@novia713: Thanks for your contribution! We'll take a look at it as soon as possible!

I can take care of writing the Twig template to render an HTML to render a PDF.

The branch has some conflicts, can you please try to sync with latest DC changes.

novia713:
merged! :smile: 

:skull: ¡ IMPORTANT ! :skull: 
I reverted the composer.json & composer.lock to master version.
in order my code to work it needs ```composer require knplabs/knp-snappy``` to be done!
otherwise this code won't work! :skull: 

i've tried to require it, but composer blindly updates all the packages and i don't know the comsequences of that! so fearing to broke something, i left unmodified the composer files ...

novia713:
merged latest changes & ready to rock!

from 0c42700 are translations

novia713:
a0aa494 includes **module version number** when listing available modules with `module:debug` command
seleccion_504

Issue #2065, opened 2016-03-26
[module:download] options


Saphyel:
I think this command needs more improvements (also in chain files)
I want this command be able to choose the version I want to use in a readable way as parameter and be able to choose where should be download (maybe I want it in sites/ or in one profiles/ )

jmolivas:
@Saphyel based on your sugesstion we should add new options:

### To select version: 
* `--version`: To enter value like `1.0-beta5`, 1.0-dev` (we can default to latest version)

### To decide where to download:
1) `--directory`: To set path (if no value entered default to `module/contrib`
2) `--type`: To set if module/profile (default to module if no value and set path)

jmolivas:
@Saphyel We cannot use `--version` as option since a global `--version` option is built-in with the Application class.

Saphyel:
@jmolivas what about use the term: `release` or `branch` ?

I prefer `release` tbh

PR #2124, opened 2016-04-11
[module:download] path option

Saphyel:
For more info #2065 

This is the first pull request for allow more options when you download a module

## Current status
- [x] Select the path


Saphyel:
@darol100 I was thinking that maybe we can split this PR into 2, so this can be merge and we deal later with `latest` & `version/release` option

darol100:
@Saphyel , I think that sounds like a great idea.

Saphyel:
@darol100 ready to test/merge

jmolivas:
@Saphyel Thanks for your contribution! We'll take a look at it as soon as possible!

config/translations/en/module.download.yml
```diff
@@ -3,6 +3,9 @@ arguments:
     module: 'Module or modules to be enabled should be separated by a space'
 options:
     latest: 'Default to download most recent version'
+    module-path: 'The path of the contrib project'
```
darol100: Can you change all module-path to just path ? We already are inside of the module it seem redundant.

Saphyel: @darol100 in fact I thought when I started to use the same rather than duplicate... because you/we ask for the same thing.

darol100: @Saphyel , That is minor improvement that we recently started to do whenever we do new enhancement, which is cleaning the redundancy between names.

Hopefully for the stable release we could have all command with clean names.

Issue #2111, opened 2016-04-07
[generate:form:config] Fatal error: Call to undefined function 5() in vendor/symfony/console/Helper/SymfonyQuestionHelper.php


boobaa:
When using the generate:form:config command in 8.1.x and v0.11.1:
```
$ drupal generate:form:config
 Enter the module name [mymodule]:
 > mymodule

 Enter the Form Class name [DefaultForm]:
 >

 Enter the Form id [default_form]:
 >


 Do you want to load services from the container (yes/no) [no]:
 >

 Do you want to generate a form structure? (yes/no) [yes]:
 >

 Type [ ]:
 > select

 Input label:
 > Foo

 Input machine name [foo]:
 >

 Size of multiselect box (in lines) [ ]:
 >
PHP Fatal error:  Call to undefined function 5() in /Users/boobaa/drupal/d8/vendor/symfony/console/Helper/SymfonyQuestionHelper.php
PHP Stack trace:
PHP   1. {main}() /Users/boobaa/bin/drupal:0
PHP   2. require() /Users/boobaa/bin/drupal:10
PHP   3. require() phar:///Users/boobaa/bin/drupal/bin/drupal:3
PHP   4. Symfony\Component\Console\Application->run() phar:///Users/boobaa/bin/drupal/bin/drupal.php:97
PHP   5. Drupal\Console\Application->doRun() phar:///Users/boobaa/bin/drupal/vendor/symfony/console/Application.php:120
PHP   6. Symfony\Component\Console\Application->doRun() phar:///Users/boobaa/bin/drupal/src/Application.php:282
PHP   7. Symfony\Component\Console\Application->doRunCommand() phar:///Users/boobaa/bin/drupal/vendor/symfony/console/Application.php:189
PHP   8. Symfony\Component\Console\Command\Command->run() phar:///Users/boobaa/bin/drupal/vendor/symfony/console/Application.php:846
PHP   9. Drupal\Console\Command\Generate\FormCommand->interact() phar:///Users/boobaa/bin/drupal/vendor/symfony/console/Command/Command.php:241
PHP  10. Drupal\Console\Command\Generate\FormCommand->formQuestion() phar:///Users/boobaa/bin/drupal/src/Command/Generate/FormCommand.php:156
PHP  11. Drupal\Console\Style\DrupalStyle->askEmpty() phar:///Users/boobaa/bin/drupal/src/Command/FormTrait.php:112
[...]

Issue #1885, opened 2016-02-08
[generate:plugin:field] fails with error


yobottehg:
i wanted to generate a field and the following error happens in the step "Enter the default field widget of the field type plugin"

PHP Fatal error:  Call to undefined function contact_form_field_widget() in phar:///usr/local/bin/drupal/vendor/symfony/console/Helper/SymfonyQuestionHelper.php on line 36

Where contact_form_field_widget is the widget i entered earlier.
The same error happens if i leave this question empty

adriangordon1231:
I am currently having the exact same error (Drupal Console version 0.10.11)

dagomar:
Confirmed, I have the same problem :'(

greg606:
ok, any progress on it?

darol100:
This seem the be fixed already. Closing this issue.

PR #2139, opened 2016-04-13
Fix "Call to undefined function xy()" issue in generate:plugin:field command.  #1885

mxr576:
Same askEmpty() issue as in #2111. Bug report: #1885


jmolivas:
@mxr576 Thanks for your contribution! We'll take a look at it as soon as possible!

jmolivas:
This makes me think we may need to fix the `askEmpty` method.

mxr576:
At least the documentation of the method, which I've already done in my previous pull request. Someone misused this method multiple times, because he/she did not know that its second parameter should be a callable.

jmolivas:
True because must be the validator.

jmolivas:
@mxr576 PR was merged. This will be included on next release, Thanks.

PR #2163, opened 2016-04-23
Improving .gitignore to support Windows, OSX, Linux, and PHPStorm…

darol100:
### Problem / Motivation
We will never want to commit anything that is related to the OS or IDE. 

### Solution
 - I have updated the .gitignore to ignore Windows, OSX, Linux, Netbeans and PHPStorm related files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,118 @@
 
 # Drupal
 /core
+
+### NetBeans ###
+nbproject/private/
+build/
+nbbuild/
+dist/
+nbdist/
+nbactions.xml
+.nb-gradle/
+
+### OSX ###
+.DS_Store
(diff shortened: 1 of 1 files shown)


jmolivas:
@darol100 Any user-specific declaration should not appear in a project .gitignore.

Consider adding it to the user global .gitignore instead.

More info:
#### Create a global .gitignore
https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

#### Ignore Symfony2 PHPStorm files like a pro with global ignoring
http://bit.ly/user-specific-gitignore

We should probably add this to the documentation.

darol100:
@jmolivas Closing this PR then.

Issue #1167, opened 2015-11-09
ContainerAwareCommand needs a getEntityTypeManager() method


bojanz:
The entity manager has been deprecated and split into 11 different services: https://www.drupal.org/node/2549139
The class that holds the getStorage() method is now the EntityTypeManager.

ContainerAwareCommand should have a getEntityTypeManager() method (core is introducing the same change to ControllerBase, forms, etc) and it should use it instead of getEntityManager().
getEntityManager() should be marked as deprecated.

darol100:
@bojanz , Closing this issue because we have another issue where we have done some of this work. https://github.com/hechoendrupal/DrupalConsole/issues/1875


Also, I have send a PR to fix this issue - https://github.com/hechoendrupal/DrupalConsole/pull/2175

darol100:
Reopening this because this issue was for the Drupal Core and the another one generate code.

PR #2175, opened 2016-04-25
Deprecating getEntityManager for entityTypeManager

darol100:
Fixes https://github.com/hechoendrupal/DrupalConsole/issues/1167

API: https://api.drupal.org/api/drupal/core%21lib%21Drupal.php/function/Drupal%3A%3AentityManager/8.2.x

We are deprecating  getEntityManager for entityTypeManager and we need to test every single command to make sure that they still to work like expected.

### How to test this PR ?
- We need to compare few command from master (or stable) and against this PR.
 - If the command is broken or have a different results please reported in this PR.
 - If the command work as expected let us know that you test it and everything work as expected. 

Here is a list of commands that needs to be tested:
- [ ] config:export:content:type
- [ ] config:export:single
- [ ] config:export:view
- [ ] database:log:debug
- [ ] generate:plugin:block
- [ ] generate:plugin:condition
- [ ] locale:language:delete
- [ ] migrate:execute
- [ ] user:debug
- [ ] user:delete
- [ ] user:login:url
- [ ] views:debug
- [ ] views:disable
- [ ] views:enable
- [ ] view:plugin:debug


src/Command/Config/ExportContentTypeCommand.php
```diff
@@ -65,7 +65,7 @@ protected function interact(InputInterface $input, OutputInterface $output)
         // --content-type argument
         $contentType = $input->getArgument('content-type');
         if (!$contentType) {
-            $entity_manager = $this->getEntityManager();
+            $entity_manager = $this->entityTypeManager();
```
naveenvalecha: better to rename this local variable. probably $entitytype_manager

src/Command/Generate/PluginConditionCommand.php
```diff
@@ -97,7 +97,7 @@ protected function interact(InputInterface $input, OutputInterface $output)
     {
         $io = new DrupalStyle($input, $output);
 
-        $entity_manager = $this->getEntityManager();
+        $entity_manager = $this->entityTypeManager();
```
naveenvalecha: we need to use entity_type.repository service instead of EntityTypeManager

src/Command/Config/ExportContentTypeCommand.php
jmolivas: based on PSR-2 coding style this should be `entityTypeManager `

src/Command/Generate/PluginConditionCommand.php
darol100: I will look at it tonight.

src/Command/Config/ExportContentTypeCommand.php
```diff
@@ -65,8 +65,8 @@ protected function interact(InputInterface $input, OutputInterface $output)
         // --content-type argument
         $contentType = $input->getArgument('content-type');
         if (!$contentType) {
-            $entity_manager = $this->getEntityManager();
-            $bundles_entities = $entity_manager->getStorage('node_type')->loadMultiple();
+            $entityTypeManager = $this->getService('entity_type.manager');
```
naveenvalecha: instead of getService method . use $this->entityTypeManager() method of ContainerAwareCommand class.
This change need at all other places.

src/Command/Generate/PluginConditionCommand.php
```diff
@@ -97,9 +97,9 @@ protected function interact(InputInterface $input, OutputInterface $output)
     {
         $io = new DrupalStyle($input, $output);
 
-        $entity_manager = $this->getEntityManager();
+        $entityTypeRepository = $this->getService('entity_type.repository');
```
naveenvalecha: @jmolivas can we add a new method entityTypeRepository() to ContainerAwareCommand class ? are we maintaining Backward compatibility ?

darol100: @naveenvalecha At some point (probably no anytime soon) we are going to stop adding service at the ContainerAware because it can be a mess to maintain.

We  do not have a stable release yet, so we are not maintaining any backward compatibility yet.

naveenvalecha: fine we can rely on the drupal core services name b/c they are not going to remove before 9.x. let's see if @jmolivas would be happy here.

jmolivas: @naveenvalecha  I like this approach better, no public functions to maintain and keep once @stable is tagged.

Issue #2167, opened 2016-04-24
[database:dump] Should not show clearly the mysql user password


novia713:
![seleccion_499](https://cloud.githubusercontent.com/assets/423279/14769023/b8fdfca6-0a4f-11e6-811f-e00229c10ede.png)

novia713:
this was already merged

PR #2182, opened 2016-04-28
[database:dump] Replace the password with asteriks

novia713:
:cyclone: #2167 :cyclone: if learning is on, we replace the password with asteriks in executed command info message


novia713:
seleccion_506

jmolivas:
There are diff ways to see the password should we hide password no those commands

* DrupalConsole commands:
  * `drupal site:status`
  * `drupal database:connect`

* Direct user interaction (nothing we can do here)
* `cat sites/default/settings.php`

novia713:
i hided the password form «site:status» & «database:connect» but we can do nothing about a «cat», «more», etc ... and obviously also not for user editing the settings.php

PR #2184, opened 2016-04-29
Fix exception when php settings were being checked

acbramley:
When settings were being checked in Application.php the following error was occuring due to the mismatch in command names.

`[Symfony\Component\Console\Exception\CommandNotFoundException]  
  Command "settings:check" is not defined.                        
  Did you mean one of these?                                      
      settings:set                                                
      settings:debug `


jmolivas:
@acbramley nice catch. Thanks for the PR

PR #2201, opened 2016-05-03
[console] Use the git commit ID for the version

opdavies:
This uses the git tag for the version, and adds a commit ID if you are using a version that has commits since the last tag.

This should make debugging easier as it's much clearer which version a user is running, and also there's no need to manually increase the version number each time.

This results in output like this:

> 10:41 DrupalConsole (use-git-version) php drupal.phar --version
Drupal Console version 1.0.0-alpha2-1-g3481b8b

### DIFF
--- a/box.json
+++ b/box.json
@@ -25,5 +25,6 @@
     "compression": "GZ",
     "main": "bin/drupal",
     "output": "drupal.phar",
-    "stub": true
+    "stub": true,
+    "git-version": "git_version"
 }
--- a/src/Application.php
+++ b/src/Application.php
@@ -30,7 +30,7 @@ class Application extends BaseApplication
     /**
      * @var string
      */
-    const VERSION = '1.0.0-alpha2';
+    const VERSION = '@git_version@';
 
     /**
      * @var string


darol100:
👍  for this....

jmolivas:
Not sure how this may affect the selg-update command

opdavies:
@jmolivas It doesn't affect it on the drupal-vm-generator project. It still works as expected.

enzolutions:
I tested the changes and looks good, but I think will be a little confusing for regulars users, will be better is the git hash or last commit is presented as extra information.

opdavies:
In most scenarios, I'd imagine that regular users would download a stable tagged version, in which case that's what they would see - the same as they would now.

enzolutions:
Uhh @jmolivas how we will differenciate the phar file create for release that the box created for dev or from source as @opdavies mentioned?

Saphyel:
With Github you have the checkbox "This is a pre-release" but I never used it.

jmolivas:
I tried this approach at the Launcher, since is the project distributed as a phar.

I agree this is helpful for debugging, but I think the hash added the commit could be confusing for users.

Issue #2205, opened 2016-05-03
[generate:form:config] Single quote in title or description results in parse error at module install


dabito:
# Problem/Motivation
If the title or description captured during the generation process for the form contain any single quotes, then the generated code contains a syntax error.

# Steps to reproduce
* drupal generate:form:config
* When asked for title include a ' in your description like: David's module
* When asked for the description include a ' in your description like. This is David's module config form
* Finish generation
* Enable the module containing the form

# Solution
I believe escaping title/description in the twig template for form config should do the trick.

miguel303:
I was looking at the issue but I can not reproduced it 
Those were the steps:
Generate a new custom module
drupal generate:form:config
On title and description I included David's module
I Finished generation and this was the generated code :
 $form['david_s_module'] = [
      '#type' => 'textfield',
      '#title' => $this->t('David&#039;s module'),
      '#description' => $this->t('David&#039;s module'),
      '#maxlength' => 64,
      '#size' => 64,
      '#default_value' => $config->get('david_s_module'),
    ]; 
Then I just enabled the module and I was able to access the form. 

@dabito can you try again please

dabito:
Sorry @miguel303 I didn't explain myself correctly: This is happenning for the menu link title and description:

```
 Generate a menu link (yes/no) [yes]:
 > yes

 A title for the menu link [DefaultForm]:
 > Default's form

 Menu parent [system.admin_config_system]:
 >

 A description for the menu link [A description for the menu entry]:
 > Description's
```

Results in:

```
module.default_form:
  title: 'Default's form'
  route_name: module.default_form
  description: 'Description's'
  parent: system.admin_config_system
  weight: 99
```

being added to the module.links.menu.yml

miguel303:
@dabito  the PR was sent in #2217 , maybe you can try updating your  drupal console version

dabito:
Thanks @miguel303! Closing as duplicate, then.

PR #2217, opened 2016-05-06
[generate:form:config] Single quote in title or description results in parse error at module install #2205

mauhg:
[https://github.com/hechoendrupal/DrupalConsole/issues/2205](url)

## Problem/Motivation

If the title or description captured during the generation process for the form contain any single quotes, then the generated code contains a syntax error.

## Steps to reproduce

- drupal generate:form:config
- When asked for title include a ' in your description like: David's module
- When asked for the description include a ' in your description like. This is David's module config form
- Finish generation
- Enable the module containing the form

## Proposed Solution

Escaping title/description in the twig template for form.


jmolivas:
@mauhg PR was merged. This will be included on next release, Thanks.

PR #2219, opened 2016-05-07
Read annotations defined on services commands

dmouse:
# Proposal
Move all command that has a dependency with some Drupal Module, this command now will be service.
Commands now should use annotations, so, we need to intercept the container compile.


jmolivas:
@dmouse can you re-roll this PR so we can test it and merge it.

jmolivas:
@dmouse thanks for the PR

This issues was resolved with PR https://github.com/hechoendrupal/DrupalConsole/pull/2642

PR #2239, opened 2016-05-10
Traits issue 2148

pierregermain:
I would like to get some feedback of the work done until today. Thank you.


jmolivas:
@pierregermain do you think this:
`Drupal\Console\Shared\ChainFilesTrait;`

is better than this:
`Drupal\Console\Command\Shared\ChainFilesTrait;`

Since those `Traits` are only used on Command classes.

pierregermain:
Hello @jmolivas . Thank you for all your help. I thought maybe for these cases we could use also `Drupal\Console\Shared\Command\ChainFilesTrait` so that all Traits are in the same place ... but really I do not have so much experience to know the best approach.

jmolivas:
I like 
`Drupal\Console\Command\Shared\ChainFilesTrait;` 
better than:
`Drupal\Console\Shared\Command\ChainFilesTrait`

and this because I mentioned below, those `Traits` are only used on `Command` classes.

jmolivas:
If any onther namespace have `Traits` we can use
`Drupal\Console\MY_NAMESPACE\Shared\MY_TRAIT;`

pierregermain:
I close this Request because to many changes have happened since I opened it. We will work on an other clean PR in the next days.

PR #2241, opened 2016-05-11
[translations] Add few more Indonesian translations.

loziju:


loziju:
Hmm weird that the check failed. How do I restart the test?

loziju:
OK really not sure what causes the application.yaml translation to fail test.

enzolutions:
Thanks @loziju

loziju:
Thanks @enzolutions but:

1.  What's with that Travis CI failure to complete for the application.yaml?
2.  And should we or should we not use quotes for the `yes` key in application.yaml:10 ? I see lots of translation files use quotes, but English doesn't have the quotes.

enzolutions:
Travis is having an issue with phar file, @omero will take care abou tthat later due he is right now at Drupal Con

Issue #2248, opened 2016-05-13
[SITE:NEW] site:new gets only stable releases


novia713:
add an option for downloading dev releases

i would add a «--dev» option for showing only dev releases

novia713:
i did a PR in https://github.com/hechoendrupal/DrupalConsole/pull/2249

jmolivas:
PR was merged

PR #2249, opened 2016-05-13
#2248 --dev option added to site:new for downloading dev releases

novia713:


jmolivas:
@novia713 Thanks for the PR.

Not sure if `--dev` could be a little confusing since versions could be `dev`, `alpha` or `beta` and `rc`

How about using `--unstable` instead?

novia713:
changed!

Saphyel:
Maybe is better follow the module approach? because could a bit confuse with only "unstable"...
e.g.
8.1.0-beta2
8.1.0-rc1
8.1.x
8.1.0
which one do you want to download with --unstable?

novia713:
@Saphyel the current status is that site:new only allow you to install stable releases. given that a --unstable option could posibilitate you download a non-stable version.
i think this would be better than actually the situation on «module» command. Module command presents you any version. This is worse, in my opinion.
But context is quite different in both cases. Actually, there is no stable version of modules for D8 in most of the cases, because of that, you get all versions in «module» command. The case of downloading drupal is totally different.
In my opinion, if modules situation were stable, an approach «stable/unstable modules to choose» would be quite better. But now, it is not the case.

jmolivas:
@Saphyel  I take a screen show to make it more clear.
site-new-unstable

jmolivas:
@novia713  PR was merged. This will be included on next release, Thanks.

Issue #2265, opened 2016-05-16
[module:uninstall] trying to reinstall module in a chain file, module:uninstall/install fails


novia713:
i have the following chain file
```
commands:
  - command: module:uninstall
    arguments:
      module:
        - '%{{module}}'
    options:
      verbose
  - command: module:install
    arguments:
      module:
        - '%{{module}}'
    options:
      verbose
```

i execute it with the following command:
```drupal chain --file="~/.console/chain/reinstall.yml" --placeholder="module:chosen"```

my idea is to uninstall & then install a given module.
strangely, it works if i do the operations manually, but it fails when using the chain file, as seen in the following screenshot:

![seleccion_571](https://cloud.githubusercontent.com/assets/423279/15290610/eca22b64-1b79-11e6-8e58-5a7886b96a84.png)

is the placeholder in .yml file wrong?

**Comment by novia713** [comment]
well, finally got it working.
![seleccion_574](https://cloud.githubusercontent.com/assets/423279/15299033/cd6e56ea-1ba2-11e6-8422-a5c228c96913.png)


seems «uninstall« expects a string and «install» expects an array.

**i firmly suggest «uninstall» should accept an array**

i leave here my chain file for other people interested on this.

```
commands:
  - command: module:uninstall
    arguments:
      module: '%{{module}}'
    options:
      verbose
  - command: module:install
    arguments:
      module:
      - '%{{module}}'
    options:
      verbose
```

jmolivas:
@novia713 +1 on accept and array of elements

novia713:
:sweat_smile: hot PR at https://github.com/hechoendrupal/DrupalConsole/pull/2268

jmolivas:
PR #2268

PR #2268, opened 2016-05-17
2265 - [module:uninstall] modules names as array when argument

novia713:
https://github.com/hechoendrupal/DrupalConsole/issues/2265

«module:uninstall» now accepts an array of arguments as module names to uninstall 

seleccion_578

also, if no module names given, it asks for them in the same way «module:install» does it.

seleccion_577

for accomplishing this, i've modified **src/Command/ProjectDownloadTrait.php**
adding a new method «_modulesUninstallQuestion()_» which suggest actually installed modules (to uninstall)


novia713:
synced & ready to rock!

Issue #1767, opened 2016-01-21
[generate:entity:content] hook_theme not generated for entities without bundles


HaleFx:
Looks like it gets generated for content entities with bundles, but not for entities without bundles. The entity won't run through the twig template without it.

cosmicdreams:
@HaleFx Yes, this feature was originally made for traditional content types.  Are you expecting it to work for custom entities' bundles?

At one point it was considered calling this feature generate:entity:bundle.  

What do you think this feature should be doing that it isn't?

HaleFx:
If I run generate:entity:content and say "no" to the bundle option, the ENTITY_TYPE.page.inc and ENTITY_TYPE.html.twig files are both created, but hook_theme is not created in the module file (actually, no MODULE.module file is created at all), so neither one is used. If you say "yes" to the bundle option, hook_theme is created (although it looks like right now the MODULE.module file is missing an opening `<?php`), so both of the above files are used. I think that hook_theme (and the .module file) should be added in both situations since the files are created and can be useful either way.

Does that make sense?

cosmicdreams:
Yes,  I don't recall coding it that way but know the place to look.  ETA: Wednesday.

frega:
@cosmicdreams @HaleFx that sounds like a regression or possibly a missing feature, not sure which. I certainly see the point to add an appropriate theme function. 

IIRC there's a small issue relating to hook_theme-entries because DrupalConsole cannot "modify" existent hook_theme-implementations so you end up w/ a new hook_theme-implementations.

PR #2272, opened 2016-05-17
Remove custom add-form and add-page functionality that is now provided by core

bojanz:
Still untested.

I think we should go back to generating the collections route in the routing.yml file, the route providers are for dynamical use cases, which is not ours. But that can happen in a different PR.

Note that entity-content.theme.php.twig and entity-content-with-bundle.theme.php.twig are now identical. Unsure how to modify EntityContentGenerator accordingly. I think my confusion comes from #1767 , we're not rendering entity-content.theme.php.twig into a file at all.


jmolivas:
@bojanz `entity-content.theme.php.twig` is not rendered is only used to output a message at `EntityContentGenerator`

```
$content = $this->getRenderHelper()->render(
            'module/src/Entity/entity-content.theme.php.twig',
            $parameters
        );

        if ($this->isLearning()) {
            $this->io->commentBlock(
                [
                    'Add this to your hook_theme:',
                    $content
                ]
            );
        }
```

jmolivas:
@bojanz I created a PR https://github.com/hechoendrupal/DrupalConsole/pull/2299 to `include` and reuse the template `entity-content.theme.php.twig` within `entity-content-with-bundle.theme.php.twig`

Issue #2294, opened 2016-05-19
[composer:XXX] add --composer option for debug, check & update modules managed with Composer


novia713:
:cyclone:  _**drupal check**_ should be aware of Composer and warning it if not installed

:cyclone:  _**module:debug**_ should have a _--composer_ option for showing modules downloaded & managed with Composer

:cyclone:  should be a _**module:update --composer**_ for updating all o a given module with Composer

i don't know how to handle all this, perhaps adding a _--composer_ option to each already existent single command would be more coherent with the general behaviour of DC ...

or perhaps would be fine to add a **composer command** with

- _drupal composer:check_
- _drupal composer:debug_
- _drupal composer:update_

in this last case i would refactor the «_module:uninstall --composer_» & «_module:download --composer_»
to
- _drupal composer:remove_
- _drupal composer:install_

***
i think we need a brainstorm on how DC will manage with Composer
- [ ] installing the site
- [ ] install modules
- [ ] remove modules
- [ ] update modules
- [ ] update the core

and how to offer this (options) to the user, how to show him this info and how interact with him

jmolivas:
I rather like to have a `--composer` option than renaming commands.

novia713:
first aproximation to «_module:update --composer_» on #2304 
for now, it does nothing without _--composer_

here is my idea on what to do without _--composer:_
* module:update (without _--composer_)
  * is the module installed?
    * NO
      * install?
        * with composer or without composer?
    * YES
      * check number version installed
      * get releases
      * show as options higher releases than installed
      * module:uninstall actual_module
      * module:install new_release
        * with composer or without it?

novia713:
#2305 related

PR #2304, opened 2016-05-20
first aproximation to module:update --composer

novia713:
first aproximation to «module:update --composer»
for now, it does nothing without --composer


novia713:
#2305 related
#2294 related

PR #2307, opened 2016-05-20
1st try to show tips with showTips Event

novia713:
doesn't work the translation of the tip


jmolivas:
@novia713 I tested this and it works out-of-the-box 

I do not have to change anything it just works and this is how it looks like:

novia713:
@jmolivas i think the code already mostly works. 
this issue needs sombody writes down a «tips» array in translation YML files of each command.
surely there are people with a knowing of DrupalConsole deeper than me, so i'm not the right person for writing those tips.

jmolivas:
@novia713 We like the idea but let's try something different

- Instead of adding a new `tips` key to the translation file, use instead the `examples` key on the translation file we have support for that and is added on the [config:override](https://github.com/hechoendrupal/DrupalConsole/blob/master/config/translations/en/config.override.yml#L16).
```
examples:
    - description: 'Set the Contact module flood limit to 10.'
      execution: 'drupal config:override contact.settings flood.limit 10'
```

- To allow the user to stop seeing this output we should add a new key to the `config.yml`. 
```
applications:
  default:
    global:
      options:
        examples: true
```

jmolivas:
@novia713  Forgot my last comment.

PR was merged. This will be included on next release, Thanks.

PR #2312, opened 2016-05-22
[update:execute] Provide distinct placeholder to fix #2293

### LINKED ISSUE #2293: Tokens in a translation sentence need to be distinct
I'm doing the translation for Chinese Traditional. We may have different order in the sentence to express the idea.

In the update.execute.yml, there are same token in a sentence, and the first one is always the number while the second one is the module name.

`executing-update: 'Executing update function %s of module %s'`
`// Executing update function 8001 of module sample`

Is it possible to make them identical?

amouro:
Replace sprintf with SafeMarkup::format.
Updated translation for en and zh_hant are also included.


jmolivas:
@amouro Updating this will make the rest of the languages stop from working since the rest of the use the `%s`

We are using the symfony translation component not the Drupal component.

Still is not clear to me why using `%s` is not working for you?

amouro:
I include the SafeMarkup because I see it's also used in user:login:url
Since the `sprintf` is less semantic, I want to propose a change as user:login:url does.

Because of the structure in different languages are vary, the order of subject and predicate may be different as well. It should be fine if a sentence only has one placeholder `%s`, but two of them may cause problem.

With `sprintf`, I have to use `%1$s` instead of `%s` to swap the placeholder in the translation sentence by using the n$ position identifier. However in original English translation file, there is only same `%s` in the sentence, if the contributor has to translate it in different order, they could make mistake. 

For example, I also provide a fix to zh_hans to the same file.
https://github.com/hechoendrupal/DrupalConsole/pull/2312/commits/23eebb52e35bd3c08a691b4a439b3d8787075429

so this leads the following thoughts and things to be discussed

1. Make it more semantic as user.login.url.yml does. 
2. If the standard is not to including Drupal component, we have to to change user.login.url.yml instead.
3. How many potential cases we may need to rearrange the placeholder, see if this is necessary for other language as well

ps. I'm not sure if it's appropriate to propose this with PR. Is there a better way that I can suggest this with my own code without PR for review?

jmolivas:
@amouro thanks for the detailed answer and clarification.

Seems like adding named placeholders will solve the issue about the position of placeholders. But we need to make sure if we change the code loading the placeholders we also replicate the change and add the placeholder on every affected language file.

Also seems like `SafeMarkup` is deprecated:
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21SafeMarkup.php/class/SafeMarkup/8.2.x

Any ideas welcome: @omero @enzolutions @darol100

darol100:
Here is the changelog that deprecated SafeMarkup::format - https://www.drupal.org/node/2549395.

We need to investigate a possible solution that might work with all languages.

enzolutions:
Hi @amouro 

Sorry for my late response.

Firstly, thanks for taking your time to made a proposal for the issue you found, that is the attitude.

About the translation, we prefer to don't have a strong link with Drupal in translation, because some commands does run as stand-alone without requiring a Drupal installation and we can not depend on Drupal to be booted.

After checking this PR the usage of sprint function feature "Argument swapping" seems like could be the solution for your language, This looks the be the solution, that is better because does not require a change in the current code and could also solve the issue about the order according to each translation.

We will add an entry in our documentation to add instructions about how to use in other translation showing how you used and a link to http://php.net/manual/en/function.sprintf.php 

We also going to remove the other usage you mentioned as at `user:login:url` and `user:login:url` command, thanks for pointing out that issue.

So, please update your Pull Request according to use the  Argument swapping the one you actually used at https://github.com/hechoendrupal/DrupalConsole/pull/2312/commits/

amouro:
@enzolutions I agree as long as we can help people know the workaround with argument swapping. It could help providing better translation.

I've update the PR with 2 revert commit. Meanwhile, I've update the `user:login:url` command with all translation accordingly. #2317

jmolivas:
@amouro Thanks for the fast reply, this PR can be closed since no file changes included.

@enzolutions Are you adding the entry on the documentation?

amouro:
@jmolivas OK, thanks for notice

PR #2315, opened 2016-05-22
module:install handles if downloaded with composer

novia713:


novia713:
- don't show D7 versions in getComposerReleases()
- removed stable versions question from module:install

jmolivas:
@novia713 PR was merged. This will be included on next release, Thanks.

PR #2334, opened 2016-05-26
user-role for adding a role to a user [working]

novia713:
comand [user:role]
the mandatory args are:
1. "add"/"remove"
2. username
3. role [machine name]

seleccion_637

//@TODO: load user by uid & email
//@TODO: discuss if username should accept an array


jmolivas:
@novia713 PR was merged. This will be included on next release, Thanks.

PR #2345, opened 2016-05-27
[generate:controller] Check if the route exits

edutrul:
- Validates path
- If router name is the same then concatenate with a random number


enzolutions:
@edutrul the PR fails due a issue with TWIG varianles, if you check the travis report you will find this error

```
  [Twig_Error_Runtime]                                                         
  Key "name" for array with keys "title, method, path" does not exist in "mod  
  ule/routing-controller.yml.twig" at line 7 
```

enzolutions:
Tested and works as expected, thanks @edutrul  for your contribution

jmolivas:
@edutrul PR was merged. This will be included on next release, Thanks.

Issue #2351, opened 2016-05-28
[module:search]


novia713:
as user i would like to search in composer repo info about modules for several reasons.
knowing the vendor of the package is one of them.

novia713:
PR in #2352 :tada:

novia713:
this no longer makes sense

PR #2352, opened 2016-05-28
[module:search] modulename --composer working [#2351]

novia713:
#2351

seleccion_658

### DIFF
--- a/config/translations/en/module.search.yml
+++ b/config/translations/en/module.search.yml
@@ -1,0 +1,10 @@
+description: 'Search modules availables for application'
+options:
+    composer: 'Search in composer repositories'
+arguments:
+    module: 'Name or names (separated by spaces) to search'
+messages:
+    only-composer: 'This command only works when passing it the option --composer'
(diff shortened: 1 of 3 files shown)


darol100:
Great work, but I think this will make confusing between debug and search. This command will provide information about each module no matter what if enable or disable (very similar to debug).

I propose to split the functionality of `module:debug` into `--enable`, which is going to only show enable modules, and this will be the default. And `--disable`, which will only show disable modules.

Then we could move this functionality another argument called `--composer`, which will provide this information. 

The end result if te user wants to show module enable or disable they could do something like   `module:debug —enable —disable —composer`. 

What you guys think ? @novia713 , @jmolivas

jmolivas:
@novia713 I like your idea about calling `package:search` and we can start using the `package` namespace for composer related commands.

novia713:
i opened a discussion thread about Composer topic in #2353

novia713:
this no longer makes sense :-/

Issue #2395, opened 2016-06-03
[module:download] & module:update doesn't work with Composer


novia713:
i've found several issues using _--composer_ option with _module:download_ and _module:update_

1. This line https://github.com/hechoendrupal/DrupalConsole/blob/master/src/Command/ProjectDownloadTrait.php#L288 afaik is wrong, because `application.composer.repositories` is an array. We need in this case `application.composer.repositories.default`. This was malforming the url repo for Composer, so i corrected this.

2. Once Composer is working, its output is not now shown. This is bad, because _Composer_ is an interactive command and may ask the user for interaction. I did several enhancements to show the Composer output in **Utils/ShellProcess.php**.

3. Regarding point 2., would be cool to _$io->info()_ the output, but i was not able to use _$io_ into **Utils/ShellProcess.php** :(

4. I firmly suggest change `application.composer.repositories.default` to `application.composer.repositories.packages` so it would be much more clear

novia713:
i also did some enhancements in md & mu string files.
PR in #2396

novia713:
i don't get why `application.composer.repositories` is an array nor why we have _default_, _drupal_ & _packagist_, but perhaps would be good idea create an option for power-users as _--repository=drupal_
for choosing one of the three, overriding the default one. 

right now in my PR i'm doing
`$this->setComposerRepositories("default");`
so one would can easly do
`$this->setComposerRepositories($input->getOption('repository'));`

jmolivas:
PR https://github.com/hechoendrupal/DrupalConsole/pull/2396 merged

jmolivas:
@novia713 Why this was re-opened ?

novia713:
i can't remember well. maybe this issue proposes many things and in that moment i were thinking in attach to this more PRs

i think we can close it already. i close it.

PR #2396, opened 2016-06-03
2395-composer-fixes

novia713:
this is for #2395


jmolivas:
@novia713 may you sync with master to test it.

novia713:
@jmolivas rerolled.
i plan to do during the weekend an option in config.yml for show/hide Composer output.
$io->info() if show
progress bar if hide
what do you think?

novia713:
seleccion_680

seleccion_679

Issue #2353, opened 2016-05-28
[package] composer commands


novia713:
I think we need discuss which **Composer's commands** we need to shortcut.
My proposal is:

|  now implemented  | proposed implementation |  composer correlation  |
|---|---|---|
|  site:new  |   |  create-project  |
|  _module:*_  |  -  |  - |
|  module:download  |  package:download  |  composer require  |
|  module:uninstall  |  package:remove  |   composer remove  |
|  module:debug  |  package:debug  |   composer show |
|  module:debug modulename |  package:debug modulename |  composer show modulename  |
|  module:search  |  package:search |  composer search  |
|  module:update  |  package:update |  composer update |
|  _theme:*_  |  -  |  - |
|  theme:download  |  package:download  |  composer require  |
|  theme:uninstall  |  package:remove  |   composer remove  |
|  theme:debug  |  package:debug  |   composer show |
|  theme:search  |  package:search |  composer search  |
|  theme:update  |  package:update |  composer update |

`site:new` i don't know how to call it, but works with composer.
`module:install` has nothing to do with composer
`module:uninstall` should manage only Drupal related things while `package:remove` should remove the package from _composer.json_

i would add:
`drupal package:outdated` to see outdated packages

so for installing a module i would do 2 steps:

1. `drupal package:download token`
2. `drupal module:install token`

and for uninstalling, the contrary:
`drupal module:uninstall token` (do things with database, etc..)
`drupal package:remove token`

what do you think, @darol100 , @jmolivas ?

darol100:
@novia713 , If all these proposed commands are wrappers to Composer, I do not think we should do that. 

Why ? I remember @bojanz mentioning that  [Composer Manager](https://www.drupal.org/project/composer_manager) was built to help the user with Composer and what it ended up doing it was actually doing is confusing the user. Because they never had an understanding of how Composer works. And they look at Composer Manager like this black box that do magic.

IMO, I think if these commands are doing the same functionality as Composer (without anything custom) we are going to end up with the same issue as Composer Manager. At this point, we should let the user use composer and not Drupal Console.

What you guys think ?
@novia713 , @jmolivas

novia713:
yeah, i get your point, @darol100. also managing Composer with DC adds more work and more code to mantain...

my first idea was that «_a only tool for managing a problem is better than two tools_», so people still don't know how to deal with Composer would be able to do all the things with DC ...

[...]

PR #2414, opened 2016-06-11
2353 theme download

novia713:
_theme:download --composer_ binds to _module:download --composer_
it downloads the theme well and _Composer_ manages where to put it

also, created a tip

> 'Pass a «true» value to «shellexec_output» key in your config.yml file to see interactively Composer''s output or «false» to hide it'

it appears in _site:new_, _module:download_ & _theme:download_

also removed a lost «`echo`» in _ShowTipsListener.php_


novia713:
this PR comes from #2353

PR #2421, opened 2016-06-14
Fix issue 2383-Form-alter

miguel303:
When we try to execute the command : drupal gfa --module="myroute" --form-id="contact_message_product_information_form" , we get the following error 
 Key "class" does not exist as the array is empty in "module/src/Form/form-alter.php.twig" at line 10

 I just added the missing keys and now the command is working properly


jmolivas:

PR #2425, opened 2016-06-17
Translate chain.debug and chain file to Punjabi

Manjitsingh:
I have translated chain.debug.yml and chain.yml file to Punjabi.

As this is my first contribution so i have translated only two files. If It gets merge and I got familiar with the process then i will get back with the large number of changes. Please review and let me know is there anything that need to be change.

thanks.


enzolutions:
Thanks @Manjitsingh

PR #2440, opened 2016-06-22
Remove unnecesary callback from template

miguel303:
I just removed this from since the function preCreate on this template does the same thing


jmolivas:

Issue #2242, opened 2016-05-11
Allow option to generate translatable content entities


SiliconMind:
When generating content entities with `generate:entity:content` the entity is always generated as non-translatable.

There should be an option asking "Is the entity type translatable?" question. Technically all that has to be done is to add proper property to entity annotation, but it would also be nice if a stub for a content translation class (inheriting from `ContentTranslationHandler`) along with translation handler annotation was also created.

jmolivas:
New release tagged https://github.com/hechoendrupal/DrupalConsole/releases/tag/1.0.0-beta3

PR #2445, opened 2016-06-23
#2242 [generate:entity:content] Allow option to generate translatable content entities

jlbellido:
This pull request try to solve #2242 

- Added a new option for set the new content entities as translatable or not.
- In case the entity is set as translatable the annotation is updated according to that and a translationHandler class is added. Nothing is done in other case.

```
$ drupal-console generate:entity:content
 Enter the module name [block_example]:
 > lopd

 Enter the class of your new content entity [DefaultEntity]:
 > ContentEntity

 Enter the name of your new content entity [content_entity]:
 > 

 Enter the label of your new content entity [Content entity]:
 > 

 Enter the base-path for the content entity routes [/admin/structure]:
 > 


 Do you want this (content) entity to have bundles (yes/no) [no]:
 > 

 Is your entity translatable (yes/no) [yes]:

```

### DIFF
--- a/Test/DataProvider/EntityContentDataProviderTrait.php
+++ b/Test/DataProvider/EntityContentDataProviderTrait.php
@@ -16,7 +16,7 @@ trait EntityContentDataProviderTrait
         $this->setUpTemporaryDirectory();
 
         return [
-          ['Foo', 'foo' . rand(), 'Bar', 'bar', 'admin/structure'],
+          ['Foo', 'foo' . rand(), 'Bar', 'bar', 'admin/structure', 'true'],
         ];
     }
 }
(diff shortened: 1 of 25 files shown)


enzolutions:
I can’t see the difference between an entity generated with support to translation to an entity without the support in Drupal UI

Please could you provide a way to test this change.

jlbellido:
This issue is a follow up of https://github.com/hechoendrupal/DrupalConsole/pull/2393 where we set the content entities as translatable always. We think that its better if we allow to the users to select if the want the entity translatable or not instead of create it always translatable.

For checking the difference between translatable and no-translatable entities in the Drupal UI you only have to follow this steps:

- Enable the language and content_translation modules
- Generate 2 content entities using the console, one translatable and another one no.
- Go to admin/config/regional/content-language
- Check that both entities are present on the list of entities.
- Notice that you only can mark the translatable entity as translatable and choose the fields that can be translated.
- This doesn't happen wih the no translatable entity.
console-entity-content-content-translation

enzolutions:
Thanks @jlbellido

enzolutions:
Works as you explain

PR #2454, opened 2016-06-24
[translation] generate:entity:content should ask for machine name (english)

ooystein:
generate:entity:content command asks for a name for the entity but the request should be for a machine name


enzolutions:
Thanks @ooystein

PR #2455, opened 2016-06-24
Validate service name has not been already taken

miguel303:


miguel303:
We are getting the following error Symfony\Component\Console\Exception\InvalidArgumentException: The helper "kernel" is not defined

which is generate by this function : 
public function getKernelHelper()
{
    return $this->getHelperSet()->get('kernel');
}

jmolivas:

PR #2474, opened 2016-06-29
[config:export] Fix PHP warnings and notices

acbramley:
ArchiveTar takes an argument in the constructor, this is causing warnings and notices during a config export command.

The variable is overwritten soon after in the conditional so is unnecessary.


jmolivas:
@acbramley PR was merged. This will be included on next release, Thanks.

PR #2477, opened 2016-06-29
[generate:plugin:block] Add default configuration to BlockPlugin template

miguel303:


enzolutions:
Thanks @miguel303 for your contributions

PR #2484, opened 2016-07-01
[behat:run] Add new command.

Saphyel:
Run behat tests from everywhere! (this is also an example of `drupal exec`)

### DIFF
--- a/config/services/behat.yml
+++ b/config/services/behat.yml
@@ -1,0 +1,5 @@
+services:
+  behat_run:
+    class: Drupal\Console\Command\Behat\RunCommand
+    tags:
+      - { name: console.command }
--- a/config/translations/en/behat.run.yml
+++ b/config/translations/en/behat.run.yml
@@ -1,5 +1,5 @@
 description: 'Run the behat tests of your application'
 error: 'Add into services.yml the parameter behat.dir with the path of your behat'
-options:
+arguments:
   filter: 'Run selected behat tests'
 
--- a/config/translations/en/behat.run.yml
+++ b/config/translations/en/behat.run.yml
@@ -1,0 +1,5 @@
+description: 'Run the behat tests of your application'
+error: 'Add into services.yml the parameter behat.dir with the path of your behat'
+options:
+  filter: 'Run selected behat tests'
+
--- a/src/Command/Behat/RunCommand.php
+++ b/src/Command/Behat/RunCommand.php
@@ -1,0 +1,47 @@
+<?php
+
+/**
+ * @file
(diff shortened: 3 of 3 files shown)


enzolutions:
Hi @Saphyel 

Thanks for your contribution. I have the following comments after test your new command

1. Error due missing services configuration
```
 [ERROR] Add into services.yml the parameter behat.dir with the path of your behat
```

2. Behat library missing? 
You need to validate if **vendor/bin/behat** is available if not report the problem 

3. Namespace
Usually, we only create a namespace when we have more than one command, please do as drupal check command for now

4. Behat debug
Would be better if you can create a behat:debug to facilitate the understanding about was test are available

Good job, I am looking forward seeing this command in our next release.

Saphyel:
@enzolutions 
1) that parameter should be added manually, because everyone have the behat tests in different places inside of the project (could be inside of webroot/tests, webroot/sites/default/tests webroot/sites/blabla.es/tests) maybe I should use a default with a warning?
2) I'll do it
3) is sensible, but I started without a clear idea how I'll end this functionality maybe is better use tests:behat:* ?

enzolutions:
Hi @Saphyel 

1) I know this is something the drupal owner have to do, but instead of to print an exception is better to capture an present a Drupal Console error message in that way we could translate to multiple languages.

2) Cool

3) Actually you are right your command my be test:behat:run and the command I suggest muest be test:behat:debug

4) I said that because I don't know if you have time to do the second command, but as I said in the comment about inside test namespace would be better.

Saphyel:
I don't mind to create more command if needed, as I said I don't have any plans for this I appreciate your tips! ^^ this is like a sporadic command lol

jmolivas:
@Saphyel @enzolutions will be this moved to `test` namespace?

enzolutions:
Yes, @Saphyel  its working on that maybe for beta5

Saphyel:
@jmolivas don't bother at the moment, I'll do it by myself. I want to use it at my local for make sense to everything rather than to random commits. it can be out of the beta.. is not urgent for me

jmolivas:
@Saphyel Worry not about random commits. I will squash them when merging.

I moved to rc-1 (wich will probably happen next week during DrupalGov).

enzolutions:
@Saphyel do you think you can complete the recommendation to include this command in our next release RC1 by August 7

Saphyel:
@enzolutions I think I'm gonna wait until Drupal 8.2 is released to continue this

enzolutions:
OK perfect

jmolivas:
@Saphyel we can probably create this as chain command.

Issue #2269, opened 2016-05-17
[site:install] Allow re-install 


Saphyel:
At the moment if you already installed the site and you want to reinstall it, you'll have to drop the database, delete the mysql config in settings.php and finally you can install it....

Could be a way to simplify this? if you are working with profiles this is a **massive problem**...

jmolivas:
👍  on this one

Saphyel:
@jmolivas thanks! in Drush you are able to install it even if there is data in the DB.
I dunno if we should create a command for restore the settings, or allow install with a _dirty DB_ or the database:drop should also restore the settings..

bradjones1:
The PR above was closed without explanation; any further thoughts on this?

bradjones1:
In Drush, you may reinstall but it appears Drupal must be first able to bootstrap. I have a workflow where it would be nice to have the installer determine the database config from the existing `settings.php` file, while not requiring the database have any data/tables.

jmolivas:
@bradjones1 this is a must have feature for the project I can test what @Saphyel was working on and resurrect the PR.

Saphyel:
Well I closed it myself because I'm not longer working on it... If you have an idea @bradjones1 feel free to overtake it.

jmolivas:
@bradjones1 @Saphyel this is done and will be included on upcoming release. 

This should be executed as 
```
drupal site:install --force --no-interaction
```

![drupal-site-install-force](https://cloud.githubusercontent.com/assets/366275/20788200/9b337e40-b764-11e6-847e-8e62c951cef5.png)

jmolivas:
PR => https://github.com/hechoendrupal/DrupalConsole/pull/2984

nvaken:
This seems merged? Is it not? Close?

bradjones1:
@nvaken Yes it is and I've been using it per above, though I don't have permissions to close. Looks done.

PR #2485, opened 2016-07-02
[site:install] Allow re-install

Saphyel:
### Description
At the moment site:install only works on fresh sites, after that is painful execute this command.

With drush you can execute site install as many times as you want and you don't even need create the table (brilliant for lazy people/beginners) I think this is the normal behavior, when you execute site:install you want to install it, check if the table exist, or if is a new site, etc...
For check all that other things, you can use a database user without permissions, or execute site:status/debug, etc...

Maybe we can keep this behavior with the parameter `--safe-install` or something like that?
Issue: #2269 

### Additional info

Maybe requires big changes like update `src/Helper/KernelHerlper.php`

### DIFF
--- a/src/Command/Site/InstallCommand.php
+++ b/src/Command/Site/InstallCommand.php
@@ -29,93 +29,97 @@ class InstallCommand extends Command
     protected function configure()
     {
         $this
-            ->setName('site:install')
-            ->setDescription($this->trans('commands.site.install.description'))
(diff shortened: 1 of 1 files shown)


jmolivas:
@Saphyel do you mind to instructions or steps to test this PR?

jmolivas:
@Saphyel do you have any plans to keep working on this?

PR #2493, opened 2016-07-04
[migrate:*] Updated commands to process migrations

miguel303:
List of changes: 

- Now we are using plugins to get a  list of migrations ids 
I removed the method getMigrations() from src/Command/ContainerAwareCommand.php and I moved it to a new Trait called MigrationTraid. 

  We load migrations using this:
 `$plugin_manager = $this->getDrupalService('plugin.manager.migration');
  $all_migrations = $plugin_manager->createInstancesByTag($version_tag);`

- To get those migrations we must to set up the database settings, so I made some changes to command migrate:setup
 This command now only load the database configuration  , so I removed the code which generate templates. 

- In order to get the dbTypes now we are calling the function directly from the core with:
 `include_once DRUPAL_ROOT . '/core/includes/install.inc';
  return drupal_get_database_types();`

- To execute a migration  , we create instances using the plugins
`$migration_service = $this->getDrupalService('plugin.manager.migration');        
  $migration_service = $migration_service->createInstance($migration_id);`

- Note: To get a list of migrations , just run drupal migrate:debug and you are going to a get the full list , so if you want filter that list , try with drupal migrate:debug 7 or drupal migrate:debug 6

### DIFF
--- a/config/services/migrate.yml
+++ b/config/services/migrate.yml
@@ -1,5 +1,9 @@
 services:
-  migrate_setup:
-    class: Drupal\Console\Command\Migrate\SetupCommand
+  migrate_execute:
+    class: Drupal\Console\Command\Migrate\ExecuteCommand
+    tags:
+      - { name: console.command }
+  migrate_debug:
+    class: Drupal\Console\Command\Migrate\DebugCommand
     tags:
       - { name: console.command }
--- a/src/Command/ContainerAwareCommand.php
+++ b/src/Command/ContainerAwareCommand.php
@@ -36,41 +36,6 @@ abstract class ContainerAwareCommand extends Command
         return $this->getKernelHelper()->getKernel()->getContainer();
     }
 
-    /**
-     * @param bool $tag
-     * @param bool $flatList
-     *
-     * @return array list of modules
-     */
-    public function getMigrations($tag = false, $flatList = false)
-    {
-        $entityType_manager = $this->getService('entity_type.manager');
-        $migration_storage = $entityType_manager->getStorage('migration');
-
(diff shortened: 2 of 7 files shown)


enzolutions:
Thanks @miguel303 for your contribution, @willwh could you test before we release

willwh:
Hi @enzolutions, I took this for a rip. Looking at the code, this is only for setting up a migration from plugins in core.

`migrate:setup` seems to be working nicely, although I don't see a way to see which migrations have been created.

`migrate:execute` works nicely for plugins loaded from core, although, it seems there is some duplication here. I don't understand the point of having `migrate:setup` and then having this as part of the `migrate:execute` command.

As an explanation Migrate Tools handles custom migrations (executing them and rolling them back, as status) - and I think DC needs to have support for this too: http://cgit.drupalcode.org/migrate_tools/tree/migrate_tools.drush.inc#n256

Also see: http://cgit.drupalcode.org/migrate_plus/tree/src/Plugin/MigrationConfigEntityPluginManager.php

That's my initial thoughts, as I said, `migrate:execute` seems to work well, but I think this is not ideal, as *most* people running migrations will not be doing a "whole hog" migration, but much more likely to use some form of custom migration, this is particularly true when migrating from Non-Drupal sources.

Take a peek at [Custom Drupal-to-Drupal Migrations with Migrate Tools](https://drupalize.me/blog/201606/custom-drupal-drupal-migrations-migrate-tools) for an idea of how to do this with Drush.

enzolutions:
@willwh ok feel free to propose a new PR to eliminate or adjust setup.

One note is, you don't need to execute all migrations, you can execute one one migration or more passing the exact migrations ids.

PR #2497, opened 2016-07-05
Spanish 20160705

novia713:
new files & lacking texts in spanish
now spanish is almost at 99% :tada:


enzolutions:
Thanks @novia713 for your contribution

PR #2501, opened 2016-07-06
[generate:plugin:fieldwidget] Do not override field widget form element

kasperg:
*Problem/Motivation*

A generated widget will not reflect basic properties of the field e.g. display label or description.

*How to reproduce*

Run `drupal generate:plugin:fieldwidget`. Subsequent input should not matter in this regard.

*Solution*

`formElement()` accepts an `$element` argument. This contains basic properties such as title, description and whether the value is required.

In the current template this argument is overridden with an empty array. As I see it there is no reason to override the variable.


kasperg:
The build fails because of code style errors in generated code. As far as I can tell the error does not seem to be related to the change at hand.

jmolivas:
@kasperg do you mind to sync with master?

kasperg:
@jmolivas: Done.

That also fixed the code style/build errors.

enzolutions:
Thanks @kasperg for your contributions

jmolivas:
PR merged. This will be included on the next release.

PR #2506, opened 2016-07-10
Do not reset $element in formElement for FieldWidget

arshad:
The $element array has #title, #title_display and #description values. `$element = [];` removes these.


arshad:
Ah duplicate of #2501. Closing.

PR #2511, opened 2016-07-12
Update some translations to pt_br

vmartinsweb:


enzolutions:
Muito obrigrado @vmartinsweb @seregatte @ratado @cristianobento

PR #2513, opened 2016-07-13
[module:install] Respect hook_requirements

miguel303:
Check module's requirements , if the module has a requirement won't be installed


jmolivas:

Issue #2389, opened 2016-06-02
Feature Request: Devel dumper change command


twhiston:
Proposed feature: command to switch the devel dumper on the site

Use Case: Its pretty easy to run out of memory with kint debugging and sometimes its useful to be able to quickly switch between dumpers

Status: Command available in https://github.com/twhiston/DrupalConsole/tree/devel-dumper

enzolutions:
Command will be restored at release 1.0.0-rc10

PR #2529, opened 2016-07-19
#2389 [database:log:poll] Add a command to poll the db logging for dev environments 

twhiston:
Command for above feature request


enzolutions:
Hi @twhiston 

Thanks for your contribution, but looks like is duplicate with command database:log:debug, please confirm.

twhiston:
@enzolutions database:log:debug does not actively poll the database output and return any new lines every x seconds, it just prints the messages that are in the log currently. The poll command was added precisely to address this shortcoming in the debug command

enzolutions:
Uhhh could you show me a screenshot of the command in action?

Do you think is possible to create a trait to reuse some code for database:log:* commands?

twhiston:
I sure can

screen shot 2016-07-19 at 16 01 55

but it makes the most sense if you run it. Basically it keeps running, and every x seconds (set by user input or defaulting to 30s) it checks the db again to see if any new log messages are in the db and if there are it prints them to the console. As you can see the first message was printed when the command was started, the 2 errors after are printed after the default 30 seconds, because i did something on my site that caused a logging event to occur

Its absolutely possible to create a trait and use that across all the database:log:* commands. I wanted to submit this command on its own before considering any major refactoring of other commands as I dont know how you guys feel, as a group, about commands that poll and arnt just 'run and done'

enzolutions:
Don't do the Trait yet, let me review the command with @jmolivas, we are traveling today, give us couple days

twhiston:
no problem at all! If you need any more info about anything to do with the command i'm happy to discuss it with you and if you guys accept it i'll happily refactor all the database commands for better code reuse

enzolutions:
@twhiston after talking with @jmolivas, we agree to include this command in next RC1, but first, you need to refactor the code to create a trait. Let me know if you are OK with that.

twhiston:
sounds like a plan. Shall i refactor the other database:log:* commands to use the trait as well?

enzolutions:
Correct, if you can refactor the other command too, would be a perfect PR for us

twhiston:
will do! What's the deadline for rc1? A bit busy at work the next week but i'd like to get all my PR's in before the deadline

jmolivas:
@twhiston The ETA for RC-1 is Monday the 8th.

twhiston:
can you let me know if how i have started to implement this trait is how you want. I wasn't exactly sure what the proper way to manage the dependencies was now, as the current dev version of Console seems to allow it, but the one we are using on our sites doesn't have any injection at all. What is the official line on this? If you can point me in the right direction I will get this finished

jmolivas:
@twhiston I will take a look a this PR

jmolivas:
@twhiston sorry for the late reply, but tagging the RC took longer that expected do you mind to sync your PR to have it merged.

PR #2540, opened 2016-07-28
Fixes typos in variable names, throws more specific exceptions

ebeyrent:
This fixes some errors in the Multisite::NewCommand class with incorrect namespace references and variable typos.  I also changed the code to throw more specific exceptions instead of \Exception.


jmolivas:
@ebeyrent PR merged. This will be included on the next release.

Issue #2542, opened 2016-07-28
Taxonomy term delete command


twhiston:
add a command to delete all terms in a taxonomy.

enzolutions:
Solved in 1.0.0-rc6

PR #2543, opened 2016-07-28
#2542 Add taxonomy delete command.

twhiston:
Added a command to delete all terms in a taxonomy. Deletion routine is a trait so can be re-used in other taxonomy deletion situations.


twhiston:
added an additional delete all command, though very similar it is a separate command as deleting every term for every taxonomy should be an explicit command due to high risk level

src/Command/Taxonomy/TaxoDeleteAllCommand.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace Drupal\Console\Command\Taxonomy;
+
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Command\Command as BaseCommand;
```
jmolivas: @twhiston  There is no longer need to register this alias. Please update this line 
```
use Symfony\Component\Console\Command\Command as BaseCommand;
```
as this
```
+use Symfony\Component\Console\Command\Command;
```

src/Command/Taxonomy/TaxoDeleteAllCommand.php
```diff
@@ -0,0 +1,41 @@
...
+use Drupal\Console\Command\Shared\ContainerAwareCommandTrait;
+use Drupal\Console\Style\DrupalStyle;
+use Symfony\Component\Console\Input\InputArgument;
+
+/**
+ * Class TaxoDeleteCommand.
+ *
+ * @package Drupal\eco_migrate
+ */
+class TaxoDeleteAllCommand extends BaseCommand {
```
jmolivas: Update this and the other command from
```
extends BaseCommand {
```
to
```
extends Command {
```

src/Command/Taxonomy/TaxoDeleteAllCommand.php
```diff
@@ -0,0 +1,41 @@
...
+  use ContainerAwareCommandTrait;
+
+  use TermDeletionTrait;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function configure() {
+    $this
+      ->setName('taxo:delete:all')
```
jmolivas: Please set command name  as:
```
->setName('taxonomy:delete:all')
```

src/Command/Taxonomy/TaxoDeleteAllCommand.php
jmolivas: Update class name from `TaxoDeleteAllCommand` to `DeleteAllCommand` since the command is already on a directory `Taxonomy`

You can see how the status commands are names on this image:
namespace-taxonomy

jmolivas:
@twhiston and the last missing piece is:

#### Register command class as a service at `config/services/taxonomy.yml`
```
taxonomy_delete:
   class: Drupal\Console\Command\Taxonomy\DeleteCommand
    tags:
      - { name: console.command }
```

jmolivas:
@twhiston how about instead of having two commands, why not only have one command and execute as:

* `taxonomy:delete VID_VALUE` - delete all terms on taxonomy with `VID_VALUE`.
* `taxonomy:delete all` - delete terms in all taxonomies.

Having to pass `all` as argument can mitigate the high risk of removing all of the terms.

twhiston:
Sorry this took a while to do, but it should be ok now.
I made it into one command that can take all to delete all terms, i refactored as requested above and I changed the command name to `taxonomy:term:delete` to make it clear that it does not delete the taxonomy itself but only the terms therein. Hope this is ok for you

twhiston:
any update on this PR? can it be included?

jmolivas:
@twhiston Now that we have the RC releases running I will be able to check and test this PR.

jmolivas:
@twhiston Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #2546, opened 2016-07-30
#2071 - [translation] Use double quotes when %s is used it.

vijaycs85:
I just ran two find/replaces under config/translations
1. find %s and replace it as "%s"
2. find ""%s"" and replace it as "%s"

Test result (local)
screen shot 2016-07-30 at 12 45 53


enzolutions:
Thanks @vijaycs85

Issue #1991, opened 2016-03-05
Alternative exiting of form creation


jmevalentin:
Right now, it seems that it will cycle through creating new fields and exits only if you don't enter a value at the new field, but it never tells you that it's adding a new field,  it just goes back to asking for a "Type". Alternatively, you should have the question, "Do you want to add a new field: yes/no [no]".

jmolivas:
We can probably try something like the `doctrine:generate:entity` command

![doctrine-generate-entity-example](https://cloud.githubusercontent.com/assets/366275/16905307/bc37d860-4c58-11e6-85f5-fac3bc453a46.png)

LOBsTerr:
@jmolivas PR https://github.com/hechoendrupal/drupal-console/pull/3747
Translations PR https://github.com/hechoendrupal/drupal-console-en/pull/160

![screenshot from 2018-01-26 10-52-27](https://user-images.githubusercontent.com/537500/35434359-29572a6e-0287-11e8-8895-6348468e1979.png)

Feel free to comment

PR #2549, opened 2016-07-30
#1991 - Alternative exiting of form creation

vijaycs85:
### Before
screen shot 2016-07-30 at 16 00 26
### After
screen shot 2016-07-30 at 16 00 12


jmolivas:
@vijaycs85 this looks good and makes adding fields more easy to understand.

enzolutions:
Thanks @vijaycs85

PR #2554, opened 2016-08-02
[translation:zh_hans] sync and update

ranqiangjun:
1. run translation:sync zh_hans
2. translate config.import.yml


enzolutions:
Thanks @ranqiangjun

PR #2559, opened 2016-08-02
Translate ja/generate.doc.*.yml

blauerberg:


enzolutions:
Thanks @blauerberg I will use to generate japanese doc

Issue #1959, opened 2016-02-23
Release 0.10.12 needs php5-curl and optionally php5-sqlite


theodorosploumis:
I just updated to version [0.10.12] (https://github.com/hechoendrupal/DrupalConsole/releases/tag/0.10.12) on Ubuntu 15.10. After that I got 1 error and 2 simple messages (recommendations) because of the new [phpcheck.yml](https://github.com/hechoendrupal/DrupalConsole/blob/master/config/dist/phpcheck.yml) file.

Error 1
```[ERROR] The extension curl is missing.```

Messages 1
```The extension sqlite3 is recommended to install.```

Message 2
```The configuration date.timezone was missing and overwritten with America/Tijuana.```

- I could solve the Error by installing the **php5-curl** extension. 
- I could solve the Message 1 by installing **php5-sqlite** extension.
- For the Message 2 I edited the ```~/.console/phpcheck.yml``` file but still no luck. My server has date.timezone set so I cannot understand why it cannot find it.

I think these should be added on the documentation somehow. Thank a lot!

**Comment by OnkelTem** [comment]
> My server has date.timezone set so I cannot understand why it cannot find it.

Have also set date.timezone in your CLI php.ini?

**Comment by theodorosploumis** [comment]
It's also in php.ini.

**Comment by OnkelTem** [comment]
I meant there are usually at minimum two php.ini-s: one for `libapache2-mod-php5` and another for CLI.

**Comment by ruodeng** [comment]
I install as the php5-fpm and nginx,  the drupal command runs as the php CLI, 
Need add the timezone on /etc/php5/cli/php.ini

**Comment by jmolivas** [comment]
You can try `php -i | grep php.ini` to check where the `php.ini` is loaded from.

**Comment by diqidoq** [comment]
This issue demonstrates that the error message isn't informing enough which timezone setting it talks about since there are many options in a server enviroment. For new Drupal 8 users it is even not sure if it is about a Drupal specific, PHP specific, Apache specific or Server specific timezone in this message.

**Comment by diqidoq** [comment]
Sadly installing *php5-sqlite* (which is the only option under php5 here) doesn't remove the message *missing sqlite3* on Debian 9. Installing php-sqlite3 installs the complete php7 stack which troubles with Drush.

**Comment by jmolivas** [comment]
@diqidoq You savoid warning message to keep appearing by executing:
```
drupal settings:set checked "true"
```

jmolivas:
[...]

PR #2566, opened 2016-08-05
#1959 Show Info that the missing configuration is a php-configuration.

zuernBernhard:
Today i stumpled upon issue #1959 with the missing time zone configuration and at first I thought a drupal configurations is meant because the message came up when I issued drupal config:diff in the terminal. Here is my idea to save other people some time.


jmolivas:
@zuernBernhard make sense, we can probably add more detailed information and mention this is from the CLI, since the ini loaded by the web-server maybe is properly configured and the issues is only from the CLI.

jmolivas:
@zuernBernhard  PR merged. This will be included on the next release.

PR #2567, opened 2016-08-05
prevent timeout of server process on windows

blauerberg:
On windows, server command is timeout every 60 seconds.

drupalconsole_timedout_windows


jmolivas:
@blauerberg PR merged. This will be included on the next release.

Issue #1863, opened 2016-02-03
Extend site:statistics to display a list of fields per content type


cosmicdreams:
In #1770 the site:statistics feature landed.  We now have a great way of seeing how many pieces of content of each kind of content exist in a Drupal 8 site.  The original intent for the creation of this command was to be able to complete a content audit more quickly than doing the archaeological dig through the site.

One additional thing that would be very insightful to discover when doing this kind of audit is to find the model of each kind of content type. To answer questions like what fields are used on each content type we'll need a way to iterate over each kind of entity, each kind of bundle, and list the fields on each.

MOAR BETTER if we can produce the outcome in an easy to read way.

In the future it would be great to generate a Build Spec like Palatir's (here: https://docs.google.com/spreadsheets/d/15htLLWLguhwiuTLg_nndQNpgWVdUMy6UaR_d1q-v6iw/edit#gid=0) I don't see anything in the code that drives this simple list of counts that make producing one of these impossible.  

We can ask Drupal about it's knowledge of the types of things it has, and thanks Drupal's discoverability, the plugin system, and the common nature of all of it's content and configuration structures, I think we can get all of the answers. The rest is just formatting (as a csv).  Twig can help us there.  

So to be clear, let's iterate our way to get there.  For now let's just focus on getting Drupal to answer specific questions about the content it's managing.  Later, we'll work on compiling the answers into a deliverable (to a team or stakeholder).

cosmicdreams:
Let's reduce the scope a bit.  For now, let's just focus on standard "node" entities.

enzolutions:
Hi @cosmicdreams 

What do yo asked is not difficult to do, actually, I have a post that explain how do that http://enzolutions.com/articles/2015/12/22/how-to-get-content-type-fields-in-drupal-8 . 

But, being honest that is not too mucho different the go to Drupal UI and see the content type definition, our intention is not to replace the UI is just do something beyond the UI in the terminal, so this feature for me is worthless the effort to be included.

The problem with another information pointed in Palantir document is related to the source of information, that information comes from Contrib module, and I not sure that we have to include that inside Drupal Console core.

cosmicdreams:
@enzolutions I hear you. Keep in mind this is a step along the way. and that the intent is to quickly answer questions about the anatomy of the structures of a Drupal site.

[...]

PR #2572, opened 2016-08-07
Listing entity field info - code adapted from http://dgo.to/field_report module

stevepurkiss:
Creating new PR for #1863 after git knowledge fail;) Will close old one.

### DIFF
--- a/config/services/field.yml
+++ b/config/services/field.yml
@@ -1,0 +1,5 @@
+services:
+    console.field_info:
+        class: Drupal\Console\Command\Field\InfoCommand
+        tags:
+            - { name: console.command }
--- a/config/translations/en/field.info.yml
+++ b/config/translations/en/field.info.yml
@@ -1,0 +1,18 @@
+description: 'View information about fields.'
+help: 'View information about fields.'
+options:
+    detailed: 'Extended output with machine names and descriptions'
+    entity: 'Restrict to a specific fieldabe entity type, for example: node, comment, taxonomy_term, shortcut, block_content, contact_message'
+    bundle: 'Restrict to a specific bundle type, for example: article'
+messages:
+    fields-none: 'No fields found'
+    entity-type: 'Fieldable entity type'
+    bundle-type: 'Bundle type'
+    not-found: 'not found'
+    in-entity-type: 'in entity type'
+    in-bundle-type: 'in bundle type'
+table:
+    header-name: 'Field label'
+    header-type: 'Field type'
+    header-desc: 'Field description'
+    header-usage: 'Also used in'
--- a/src/Command/Field/InfoCommand.php
+++ b/src/Command/Field/InfoCommand.php
@@ -1,0 +1,252 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\Console\Command\Field\InfoCommand.
+*/
+
+namespace Drupal\Console\Command\Field;
+
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Command\Command;
+use Drupal\Console\Command\Shared\ContainerAwareCommandTrait;
+use Drupal\Console\Style\DrupalStyle;
+use Drupal\field\FieldConfigInterface;
+
+/**
+ * Class InfoCommand.
+ */
+class InfoCommand extends Command
+{
+    use ContainerAwareCommandTrait;
+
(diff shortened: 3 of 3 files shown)


stevepurkiss:
So... you can now restrict `field:info` to a specific entity type, for example:

`drupal field:info --entity-type=node`

If you select one which doesn't exist, e.g. `nodes` it will display an error message. This doesn't cope for if you select a non-fieldable entity type, not sure how necessary it is to cope for that - I guess I could say entity type found but not fieldable, am parking that one for the moment as it's tea-time @purkisshq!

field-info-entity-type

The code is still very if...else etc. rather than funky functions but hey, I'm old & can always refactor :D

stevepurkiss:
...or I could just change the error message to say 'Fieldable entity type X not found'... let me know your thoughts.

stevepurkiss:
Actually I might just do it ;) brb...

stevepurkiss:
There we go.

field-info-entity-type-not-found

stevepurkiss:
Just improving my error checking code, didn't actually check whether a specific entity type was selected when checking whether it was found or not. It would only error if there were no fieldable entities anyway, but, y'know... ;)

stevepurkiss:
OK, I think this is ready for review! You can:

- select `--detailed` output which displays descriptions and machine names of entities and bundles
- select `--entity=[ENTITY]` to restrict to a specific entity type e.g. `--entity=node` (shortened but can revert to `--entity-type` if that's 'better')
- select `--bundle=[BUNDLE]` to restrict to a specific bundle type e.g. `--bundle=article` (shortened but can revert to `--bundle-type` if that's 'better')

field-info-entity-bundle

stevepurkiss:
Just added a counter so if there's no fields it displays an appropriate message, and made the current messages re no fields a bit more descriptive. Still ready to test, now looking into how I can make the code more 'current' than procedural ;) Still works though...

stevepurkiss:
Hmm... think I'm gonna leave the code as it is for the moment - there's little I can see that'll improve the computational speed by splitting up the execute function. I'm open to suggestions as realise my mind mostly sees procedural code & would like to improve that ;)

The main reason for lots of code is the ability to get a list of fieldable entities from Drupal (as opposed to a list of just fieldable node types), most of that code is from the source module http://dgo.to/field_report.

I've looked through the latest Drupal API & found [EntityTypeBundleInfoInterface](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityTypeBundleInfoInterface.php/interface/EntityTypeBundleInfoInterface/8.2.x) & tried to get it working yesterday but without luck - if this is a Good Way Forward then let me know your thoughts & will pursue that line, in the mean time I'm gonna have a break as think this is Good Enough for a 1st version of `field:info` command!

stevepurkiss:
Just fixing a standards issue, forgot to run phpcs ;)

stevepurkiss:
FYI am going through sorting out dependency injection etc., squashing & pushing as I get to incremental working stages so code *should* improve ;)

stevepurkiss:
OK, I have to focus on other stuff for a bit so leaving as-is for now, welcome any input/feedback on the code so far - have tidied up design a bit too so it's using $io->comment instead of $io->commentBlock which is a bit 'in yer face' for the actual messages output.

Hope this is ok - functionality will hopefully be useful to ppl!

stevepurkiss:
OK ignore my last comment - just noticed a big chunk of code from the original module which was not really necessary - renaming the entity / bundleParent labels, so just using the labels.

stevepurkiss:
OK, seems EntityTypeBundleInfoInterface functionality is available from the entity type manager but doesn't seem to help reduce the amount of functionality needed to get the list of fieldable entities so leaving that as is.

Will improve code if I figure anything else out, apart from that it's gonna sit here waiting review - everyone's busy on rc1 at the moment so looking forward to seeing that get released yay!

stevepurkiss:
OK, here's an updated screengrab showing a few options and a couple of errors.

You can now:

- select `--detailed` output which displays descriptions and machine names of entities and bundles
- select `--entity=[ENTITY]` to restrict to a specific entity type e.g. `--entity=node` (shortened but can revert to `--entity-type` if that's 'better')
- select `--bundle=[BUNDLE]` to restrict to a specific bundle type e.g. `--bundle=article` (shortened but can revert to `--bundle-type` if that's 'better')

field-info-output

jmolivas:
@stevepurkiss looks good to me.

jmolivas:
I will merge it and migrate to new command structure.

PR #2583, opened 2016-08-11
[WIP] Remove relocated code on drupal-console-core

jmolivas:


daften:
Am I correct when I say that the PHP minimum version is now 5.6.x instead of 5.5.9 (in the PHP check command)? Doesn't this make Drupal console stricter than Drupal itself?

jmolivas:
@daften We did that because PHP 5.5 is no longer supported 
http://php.net/supported-versions.php

I am planning to update the checker to to ask for a minimum version `5.5.9` as Drupal 8 and a suggested one `5.6`.

daften:
Aha, yes, EOL always comes quicker than expected :) Seems like a good plan, thanks for explaining!
For reference, relevant drupal.org issue: https://www.drupal.org/node/2524432

PR #2587, opened 2016-08-16
Adding twig:lint and lint:twig commands.

grasmash:


grasmash:
Hm. The twig-bridge library does require PHP 5.6+.

jmolivas:
Since this requires `php7` will be better to manage as separate package. 

You can use this project as an example https://github.com/hechoendrupal/drupal-console-extend-example to do this. Feel free to ask for help.

PR #2623, opened 2016-08-25
Issue 2621 add configexport flag

allgood2386:


src/Command/Config/ExportSingleCommand.php
```diff
@@ -70,6 +70,11 @@ protected function configure()
                 '',
                 InputOption::VALUE_OPTIONAL,
                 $this->trans('commands.config.export.single.options.optional-config')
+            )->addOption(
+                'include-uuid',
```
jmolivas: I think better to default to true.

Rename this to `--remove-uuid`

src/Command/Config/ExportSingleCommand.php
```diff
@@ -197,9 +202,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $module = $input->getOption('module');
         $configName = $input->getArgument('config-name');
         $optionalConfig = $input->getOption('optional-config');
+        $uuid = $input->getOption('uuid');
 
-        $config = $this->getConfiguration($configName);
-
+        if ($uuid) {
+            $config = $this->getConfiguration($configName, TRUE);
```
jmolivas: DrupalConsole follows PSR-2 code style.
* `TRUE` must be `true`
* `FALSE` must be `false`

allgood2386:
@jmolivas once the translation is merged in I think this is ready for review.

Issue #2593, opened 2016-08-18
[meta] Register commands in the Drupal container.


jmolivas:
### Problem/Motivation

Register Drupal Console commands as services on the Drupal container.
### Why
- Improve command discover and take advantage of dependency injection.
#### Extend from `Symfony\Component\Console\Command\Command`

Add `use` keyword if not present

```
use Symfony\Component\Console\Command\Command;
```
#### Update class to extend from `Command`

```
class AwesomeCommand extends Command
```
#### Add `CommandTrait`

Add `use` keyword if not present

```
use Drupal\Console\Command\Shared\CommandTrait;
```

If command is using 

```
Drupal\Console\Command\Shared\ContainerAwareCommandTrait;
```

Update to 

```
Drupal\Console\Command\Shared\CommandTrait;
```

Add `use` trait:

```
use CommandTrait; 
```
#### Register command class as a service at `config/services/NAMESPACE.yml`

```
console.awesome:
   class: Drupal\Console\Command\AwesomeCommand
   arguments: ['@service_id', '@other_service_id', '@another_service_id']
   tags:
      - { name: drupal.command }
```

NOTE: If command already registered make sure you remove comment definition and confirm service name is prepended with `console.`. 
#### How to know which services to inject

Any call to:

```
$this->getService('SERVICE_NAME')
$this->getDrupalService('SERVICE_NAME')
$this->get('SERVICE_NAME')
\Drupal::service('SERVICE_NAME')
\Drupal::getContainer()->get('SERVICE_NAME')
$this->getApplication()->getContainerHelper()->get('SERVICE_NAME')
```

Should be added as `argument: ['@SERVICE_NAME']` on the service definition.
#### Inject service add use, property and constructor method.

```
...
use Vendor\Package\ServiceName;
...

    /**
     * @var ServiceName
     */
    protected $serviceName; 

    /**
     * AwesomeCommand constructor.
     * @param ServiceName $serviceName
     */
    public function __construct(ServiceName $serviceName) {
        $this->serviceName = $serviceName;
        parent::__construct();
    }
```

NOTES: 
- You can find the `Vendor\Package\ServiceName` of your service by executing from your Drupal 8 site `drupal container:debug SERVICE_NAME`.
- Make sure you call the `parent::__construct();` on the added `__construct`method.
#### Some methods and classes has been relocated to new service and maybe renamed, this is the table mapping those changes.

| Current | New Service name | New Method name |
| --- | --- | --- |
[...]

PR #2633, opened 2016-08-27
2593 module namespace

novia713:
`module:install` with injection of dependencies working

just touched the code also for when using `--composer` option it checks wether if we are in "web" or "docroot" and then execute `composer` from an upper directory

i bet without `--composer` option it works well, with this option the only problem i see is when the module wasn't downloaded with composer the `exec` method of `shellprocess` should stop the run

now it looks like this
seleccion_460

anyway, i guess the `--composer` option will be removed, so this is not very important ...


novia713:
this PR comes from #2593

novia713:
`module:uninstall` reconverted is working

anyway, `module:install` is giving me the following error, any advice?
```
../vendor/bin/drupal module:install devel
PHP Fatal error:  Call to undefined method Drupal\Console\Application::getValidator() in /home/leandro/public_html/hq/vendor/drupal/console/src/Command/Shared/ProjectDownloadTrait.php on line 84
```

jmolivas:
@novia713 there is a valiator service
```
  console.validator:
    class: Drupal\Console\Utils\Validator
```

novia713:
done work in PathCommand & UpdateCommand @jmolivas 

don't kwon why is erroring at inyecting @console.shell_process
this still needs a bit of work extra

jmolivas:
Not sure if `--composer` will stay or go we should talk about this

PR #2663, opened 2016-09-06
Register theme commands as services in Drupal Container

enzolutions:


src/Command/Theme/DebugCommand.php
```diff
@@ -47,7 +73,7 @@ protected function themeList(DrupalStyle $io)
             $this->trans('commands.theme.debug.messages.version'),
         ];
 
-        $themes = $this->getService('theme_handler')->rebuildThemeData();
+        $themes = $this->themeHandler->rebuildThemeData();
         $tableRows = [];
         foreach ($themes as $themeId => $theme) {
             $status = $this->getThemeStatus($themeId);
```
jmolivas: You should pass `$theme` instead of `$themeId`
```
$status = $this->getThemeStatus($theme);
```

jmolivas:
On `InstallCommand` and `UninstallCommand.php` command you should inject the `@console.chain_queue` service and then replace this:
```
$this->get('chain_queue')->addCommand('cache:rebuild', ['cache' => 'all']);
```
with this:
```
$this->addCommand('cache:rebuild', ['cache' => 'all']);
```

PR #2680, opened 2016-09-13
Fixes #2276. Default value for uri option. Without it it's impossible to vie…

### LINKED ISSUE #2276: Provide default uri from settings.php
Currently the --uri option does not fallback to a default value.

It would be great if you could provide a default uri value in settings.php (or local.settings.php) that would be used unless --uri is provided to override it.

blazeyo:
…w messages logged during command execution in Reports UI (InvalidArgumentException).

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -200,12 +200,16 @@ class App extends BaseApplication
 
         $uri = $input->getParameterOption(['--uri', '-l']);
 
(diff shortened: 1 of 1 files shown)


blazeyo:
Fixes #2276

jmolivas:
@blazeyo PR merged. This will be included on the next release.

PR #2700, opened 2016-09-16
Add a new way to contribute

Saphyel:


README.md
```diff
@@ -81,9 +81,17 @@ drupal self-update
...
-
 For information about how to run this project for development follow instructions at [setup instructions](https://gist.github.com/jmolivas/97bbd07f328217be3564a434c5bd2618).
 
+An optional way you can execute this:
+```
+composer require --dev drupal/console:dev-master
+```
+You might have to execute too:
+```
+composer require symfony/dom-crawler:2.8 symfony:debug:2.8
```
jmolivas: Why are these required?

Saphyel: @jmolivas  I think drupal/core require the version 2.7 of both them, if you try to install with composer the dev-master of drupal/console you need at least the version 2.8 so because of that composer says "Hey there is conflict of dependencies with 2 plugins do something"

novia713: **well, firstly, master development branch already has ~2.8 versions of dom-crawler & debug as requirements.**
so, the statement `You might have to execute too` is only intended for the stable version (beta5) ....

secondly, regarding to `beta5` (which has 2.7 versions) i had no problems ...

Saphyel: Yep, but if you want to use it in the drupal composer.json (Drupal 8.3.x core for instance) is still using 2.7 so composer finds a conflicts and won't be able to install drupal/console

is very easy to test this.. In a drupal project run the command `composer require --dev drupal/console:dev-master`

pfrenssen:
Hmm I think this information is too vaguely described and too trivial to have a lot of added value in the README.

I think this is more suitable to place on an online help forum, for example as a Q&A on the Drupal stack exchange.

PR #2738, opened 2016-09-26
Adding site:import:local as a service

ibonelli:
Hi Enzo, this is the site:import:local  you helped me with. Thanks!


enzolutions:
Thanks @ibonelli for your contribution,

PR #2750, opened 2016-10-01
WIP: Use composers autoloader for path retrieval

webflo:


Saphyel:
I like the idea!

webflo:
This is blocked by the facet that DrupalConsole, drupal-console-core and drupal-console-launcher use the same PSR-4 namespace. They have to use different namespaces. 

@jmolivas Can we change drupal-console-core to Drupal\ConsoleCore and drupal-console-launcher to Drupal\ConsoleLauncher?

enzolutions:
I like the idea @webflo

jmolivas:
@webflo What is the reason for renaming the namespace, if this is the case we will need to change code in core and contrib commands, may we search for another option?

webflo:
Composer uses the namespace as array key in its PRS-4 autoload (look at an composer/autoload_psr4.php). It will only use DrupalConsole or drupal-console-core. You can't have both in one project. It fails if you try to pull both as dependency in your project.

webflo:
I think its safe to start with the launcher. This component is separate, the change is easy. We have to do more planning for the other ones.

jmolivas:
Namespaces updated
* drupal-console-core
https://github.com/hechoendrupal/drupal-console-core/blob/master/composer.json#L63

* drupal-console-launcher
https://github.com/hechoendrupal/drupal-console-launcher/blob/master/composer.json#L61

* drupal console 
https://github.com/hechoendrupal/drupal-console/pull/3051

jmolivas:
@webflo  We changes the constant definition extending `DrupalFinder` and using the `getVendorDir` value from DrupalFinder class.

https://github.com/hechoendrupal/drupal-console-core/blob/master/src/Utils/DrupalFinder.php

enzolutions:
@webflo this idea is still valid, thanks in advance for any comment in regards to it.

PR #2757, opened 2016-10-04
fix #2756: maintain maintenance mode if previously enabled

### LINKED ISSUE #2756: [update:execute] Maintenance mode is disable after execution even if previouly enabled
Maintenance mode is disable after execution even if previouly enabled.
No check is done to persist an eventual maintenance mode

smalot:


jmolivas:
@smalot Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #2791, opened 2016-10-11
Enable pass configutation to migration plugins

enzolutions:
Migration plugin with files doesn't work, because can't allocate the drupal directory webroot folder/path


enzolutions:
Thanks @me :P

me:
You're welcome!

PR #2799, opened 2016-10-12
[site:statistics] command fails when comment module is uninstalled. Fix #2744

### LINKED ISSUE #2744: [site:statistics] command fails when comment module is uninstalled.
I ran the site:statistics command on a site where the comment module was disabled/uninstalled and received the following error:
 [Drupal\Component\Plugin\Exception\PluginNotFoundException]
  The "comment" entity type does not exist.

Drupal                       8.1.8
Drupal Console version 1.0.0-rc1

daydiff:
fix failing command [site:statistics] when comment module is uninstalled

### DIFF
--- a/src/Command/Site/StatisticsCommand.php
+++ b/src/Command/Site/StatisticsCommand.php
@@ -10,6 +10,7 @@ namespace Drupal\Console\Command\Site;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Command\Command;
+use Drupal\Component\Plugin\Exception\PluginNotFoundException;
 use Drupal\Console\Command\Shared\ContainerAwareCommandTrait;
 use Drupal\Console\Style\DrupalStyle;
 use Drupal\Console\Utils\DrupalApi;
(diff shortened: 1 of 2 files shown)


src/Command/Site/StatisticsCommand.php
```diff
@@ -115,7 +116,12 @@ private function getNodeTypeCount($nodeType)
      */
     private function getCommentCount()
     {
-        $entityQuery = $this->entityQuery->get('comment')->count();
+        try {
+            $entityQuery = $this->entityQuery->get('comment')->count();
+        } catch (PluginNotFoundException $e) {
+            return 0;
```
daydiff: But it is not the value CLI application returns, but the value of comments which is 0 when "Comment" module is not installed. Btw I agree, that it would be better to not catch exception but prevent it. I can replace catch block with something like this:

```php
if (!\Drupal::moduleHandler()->moduleExists('comment')) {
    return 0;
}
```
What do you think about such changes?

daydiff: I replaced try-catch block so it is no exception now.

jmolivas: My bad, but like this better value 0 instead of `PluginNotFoundException`

jmolivas:
@daydiff Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #2802, opened 2016-10-13
Fix #2801 fix failing command [generate:module]

qinshuguo:

Issue #2801, opened 2016-10-13
[generate:module] PHP Fatal error:  Call to undefined method Drupal\Console\Command\Generate\ModuleCommand::validateModuleName() in ./vendor/drupal/console/src/Command/Generate/ModuleCommand.php on line 271


qinshuguo:
PHP Fatal error:  Call to undefined method Drupal\Console\Command\Generate\ModuleCommand::validateModuleName() in ./vendor/drupal/console/src/Command/Generate/ModuleCommand.php on line 271

PacificMorrowind:
ah I think I found the right commit message to trigger the auto-convert to PR from Issue.

PacificMorrowind:
Wait no I didn't and now I see there already was a PR for this although it only fixed the one bug. Want me to delete  my branch and redo it as a new PR  that only fixes the second bug?

jmolivas:
PR https://github.com/hechoendrupal/DrupalConsole/pull/2808 merged

PR #2808, opened 2016-10-13
Fix incorrect validation call

PacificMorrowind:
Only triggered if --machine-name is used in the module:generate call.
see also #2801 and #2802


jmolivas:
@PacificMorrowind Thanks for your contribution, your PR was merged. This will be included on the next release.

Issue #2812, opened 2016-10-15
[taxonomy:term:delete] Inject dependencies.


jmolivas:
Take advantage of DI and the service container injecting dependencies  

Service definition:
https://github.com/hechoendrupal/DrupalConsole/blob/master/config/services/drupal-console/taxonomy.yml

Command class:
https://github.com/hechoendrupal/DrupalConsole/blob/master/src/Command/Taxonomy/DeleteTermCommand.php

nidble:
I'm working on it!

nidble:
Hi @jmolivas whats do you think about my PR? Is there something not correct or wrong?

jmolivas:
Thanks for the PR. I will be reviewing during BADCamp this week.

jmolivas:
@nidble  Thanks for the PR. I will be reviewing during BADCamp this week.

LOBsTerr:
@jmolivas It can be closed

jmolivas:
Closing this one. Feel free to re-open if still happening after updating to the latest version.

PR #2828, opened 2016-10-18
[taxonomy:term:delete] Inject dependencies. (#2812)

nidble:
Inject and use dependencies in  [taxonomy:term:delete] #2812


enzolutions:
Thanks @nidble for your contributions

PR #2830, opened 2016-10-18
2821 translation migration

novia713:
`translation` commands migrated to Drupal container
--
reviewing my commits in github interface i noticed:
- accidentally commited my composer.json
- the code needs to be well formatted
(i'm still struggling with my code-editor)


enzolutions:
thanks @novia713

Issue #2831, opened 2016-10-19
generate:entity:content with revision = yes results in Fatal Error


marcusx:
`drupal generate:entity:content`
Answer `commands.generate.entity.content.questions.revisionable (yes/no) [yes]:` with yes

I get:

```
PHP Fatal error:  Call to undefined method Drupal\Console\Generator\EntityContentGenerator::getSite() in /Users/marcus/Projekte/.../vendor/drupal/console/src/Generator/EntityContentGenerator.php on line 183
```

Using Console: (1.0.0-rc6)
And: Drupal Core 8.2.1

betoquiroga:
hi,   I have the same issue but with another error when I enable the module

`[ERROR] Drupal\test1\TestEntityStorage does not implement Drupal\Core\Entity\ContentEntityStorageInterface`

jmolivas:
@marcusx @betoquiroga seems like the issue is this
https://github.com/hechoendrupal/DrupalConsole/blob/master/src/Generator/EntityContentGenerator.php#L180-L211
We need to replace this
```
$this->getSite()
```
with
``` 
$this->site
```

marcusx:
@jmolivas here a PR

PR #2832, opened 2016-10-19
[generate:entity:content] Fix undefined method error

marcusx:
fix for https://github.com/hechoendrupal/DrupalConsole/issues/2831


enzolutions:
Hi @marcusx 

Thanks for detect the error, but the change you propose doesnt fix the error.

Instead of that you must use the extension manager service as you can see below

```
$this->extensionManager->getModule($module)->
```

I am going to close this PR and fix myself, because this is a critical issue and we are about to deliver a new release in few hours and would be better if we fixed the command for the rest of community.

Again thank you for your report.

PR #2847, opened 2016-10-23
EntityContentGenearator should store *.page.inc in module root direct…

clrockwell79:
…ory, not /src

This patch puts the *.page.inc file in the module root directory, which is where hook_theme (@see entity-content.theme.php.twig) expects it to be.


jmolivas:
@clrockwell79 Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #2424, opened 2016-06-16
Period remove from generated .module file file block.

iStryker:

PR #2849, opened 2016-10-25
Update module.twig

NickWilde1990:
(Excessively minor - 1 char).
remove doubled up period (base/file.php.twig already includes a period right after the file_path block).

### DIFF
--- a/templates/module/module.twig
+++ b/templates/module/module.twig
@@ -1,6 +1,6 @@
 {% extends "base/file.php.twig" %}
 
-{% block file_path %}{{machine_name}}.module.{% endblock %}
+{% block file_path %}{{machine_name}}.module{% endblock %}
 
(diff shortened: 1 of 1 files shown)


jmolivas:
@NickWilde1990 seems like duplicated https://github.com/hechoendrupal/DrupalConsole/pull/2424

I added a comment on the other but no reply by @iStryker  yet

iStryker:
Yes sounds like a duplicate.   Sorry, I did not see your comment before.

jmolivas:
@iStryker the duplicated si the other PR.
We should make this PR to output as the code on my comment.

NickWilde1990:
ah yes, didn't see 2424, sorry.

jmolivas:
@NickWilde1990 worry not and thanks for the PR.

PR #2856, opened 2016-10-28
Fix msg for module:uninstall with dependencies.

NickWilde1990:
Currently if you run module:uninstall with f.e. comment and node, it will print this helpful message:
`[ERROR] Unable to uninstall modules "" because are required by "standard"`
this is because the module variable is null
if fixed to correctly implode the moduleList variable it'll print
`[ERROR] Unable to uninstall modules "comment, node" because are required by "standard, history, taxonomy, standard"`
but `[ERROR] Unable to uninstall modules "comment", "node" because are required by "standard, history, taxonomy, standard"` is better.

### DIFF
--- a/src/Command/Module/UninstallCommand.php
+++ b/src/Command/Module/UninstallCommand.php
@@ -173,7 +173,7 @@ class UninstallCommand extends Command
                 $io->error(
                     sprintf(
                         $this->trans('commands.module.uninstall.messages.dependents'),
-                        implode(', ', $module),
+                        implode('", "', $moduleList),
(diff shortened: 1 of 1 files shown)


NickWilde1990:
Oh couple notes: 
1. This was also giving a php warning in the drupal log: "Warning: implode(): Invalid arguments passed in Drupal…"
2. there is an undefined variable earlier but it is a separate issue really.

jmolivas:
@NickWilde1990 Nice catch. Thanks for your contribution, your PR was merged. This will be included on the next release.

Issue #2859, opened 2016-10-29
Pass «query» option to database:client to execute some determined query


novia713:
Use case:
imagine i would like to execute some query on my site like 
`DELETE FROM key_value WHERE name like 'fontawesome';`

i would pass an option "query" to _database:client_ to avoid enter the mysql console and executing directly this query, like
`../vendor/bin/drupal database:client --query="DELETE FROM key_value WHERE name like 'fontawesome';"`

i know this is dangerous and output should be clear, even a confirm message would be nice; but also this is very helpful for chain-scripts or bash-scripts

jmolivas:
@novia713 like the idea but probably this will fit better into a new command how about `database:query`

novia713:
`database:query "my-sql-statement"` working

jmolivas:
PR https://github.com/hechoendrupal/DrupalConsole/pull/2860 merged

PR #2860, opened 2016-10-29
[database:client] "query" option

novia713:
added «query» option for executing directly via option SQL sentences when using database:client
related to #2859 
//@todo: show output


novia713:
added several options to this command in this PR
seleccion_036

novia713:
bump

novia713:
@jmolivas bump

Issue #2862, opened 2016-10-30
[database:dump] add --tar option 


novia713:
would be great if `database:dump` would have a `--tar` option like `config:export`

i can work on this if the maintainers would consider it useful

jmolivas:
@novia713  I had been asked a few times about this feature.

novia713:
already merged

PR #2863, opened 2016-10-30
added --gz option to database:dump

novia713:
PR for #2862

i preferred to use native php functions for gzipping the file rather than using Drupal ArchiveTar class.
i did

``` php
file_put_contents(
  $resultFile,
    gzencode(
        file_get_contents(
            $file
            )
    )
);
```

perhaps you don't like it :upside_down_face:


enzolutions:
Thank you @novia713 RTBC

PR #2879, opened 2016-11-03
Declare data_table in annotation only if entity is translatable

blackice2999:
If you generate a entity that is not translatable dont define "data_table" annotation because of side effects with views.

See for data_table implementaion @ https://www.drupal.org/node/1498674


jmolivas:
@blackice2999 Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #2880, opened 2016-11-03
Fixes wrong template path on generate:entity:content

blackice2999:
After generate of a entity the templates files should be in MODULENAME/templates not in MODULENAME/src/templates this causes a exception on entity view.


jmolivas:
@blackice2999 Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #2900, opened 2016-11-08
[update:execute] Require correct '.install' file for each module during updates

esolitos:
As [discussed in Gitter](https://gitter.im/hechoendrupal/DrupalConsole?at=58216ff378ec59ab05479d07) in the latest rc8 seems like that `drupal update:execute all` is broken.
The reported error is 

> Call to member function getPath() on null in […]src/Command/Update/ExecuteCommand.php on line 187

I believe that the `loadLegacyFile` should be including the current install module, not the parameter as fixed in this PR.


enzolutions:
Thanks @esolitos

PR #2910, opened 2016-11-10
Rename getControllerDirectory() to getControllerPath() in Ext Mgr

### LINKED ISSUE #2881: exception 'ReflectionException' with message 'Class \Drupal\testmodule\Controller\TestModuleEntityController does not exist' with generate:entity:content and revisions enabled
It looks like when revision capabilities are utilized in generate:entity:content an exception is thrown.  I don't know enough to help with it, but the generation code lives in templates/module/src/entity-content-route-provider.php.twig and was introduced in 0dc8f715.  The offending Class code is not generated when revisions are disabled.  I've attached a rundown of the setup including the exact steps used to generate the error.  The error appears at the bottom of the document.

I'm not sure if it's something I'm doing or if it's missing code in the generate:entity:content portion of Drupal Console. :(

[Drupal Console - New Module Revisions.txt](https://github.com/hechoendrupal/DrupalConsole/files/570370/Drupal.Console.-.New.Module.Revisions.txt)

EDIT: Note, there are errors in the above text when using drupal console; "commands.yaml.unset.key.yml Unable to parse at line 2 (near "  arguments:")."  I am pretty sure those will be fixed in [2878](https://github.com/hechoendrupal/DrupalConsole/pull/2878), though the version of console used is 1.0.0-rc7

gone3uo6:
Use new name in Controller Generator. Fixes #2881

### DIFF
--- a/src/Extension/Extension.php
+++ b/src/Extension/Extension.php
@@ -27,9 +27,9 @@ class Extension extends BaseExtension
      * @param bool $fullPath
      * @return string
      */
-    public function getControllerDirectory($fullPath = false)
+    public function getControllerPath($fullPath = false)
     {
-        return $this->getSourcePath($fullPath) . '/Controller/';
+        return $this->getSourcePath($fullPath) . '/Controller';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jmolivas:
@gone3uo6 Thanks for your contribution, your PR was merged. This will be included on the next release.

Issue #2911, opened 2016-11-10
[rest:disable] Does not disable as expected


edutrul:
# Problem/Motivation

I was trying to disable a REST, but when I wanted to check it out in /admin/config/services/rest it doesn't disable as expected, it stays in enabled state.

# How to reproduce

Steps:
1. Make sure you have restui contrib module install https://www.drupal.org/project/restui
2. Go to /admin/config/services/rest) install a entity resource like for instance "content" which name is "entity:node"
3. You will see "Content" is enabled
4. Then execute "rest:disable"
5. After that type "entity:node"
6. You will see that says " [OK] Rest ID "entity:node" disabled"
7. Go back to /admin/config/services/rest
8. You will see it is not disabled.

Other details:
Drupal version: 8.2
Console version: 1.0.0-rc8

# Solution

I am going to be working on it and soon I'll have a PR asap

PR #2916, opened 2016-11-11
[rest:disable] Fix rest disable functionality

edutrul:
According to issue https://github.com/hechoendrupal/DrupalConsole/issues/2911

### DIFF
--- a/src/Command/Rest/DisableCommand.php
+++ b/src/Command/Rest/DisableCommand.php
@@ -30,13 +30,13 @@ class DisableCommand extends Command
     use RestTrait;
 
     /**
- * @var ConfigFactory  
-*/
+     * @var ConfigFactory  
+     */
     protected $configFactory;
 
     /**
- * @var ResourcePluginManager  
-*/
+     * @var ResourcePluginManager  
+     */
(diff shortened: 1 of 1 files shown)


src/Command/Rest/DisableCommand.php
```diff
@@ -96,23 +95,38 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+            $io->success(
+                sprintf(
+                    $this->trans('commands.rest.disable.messages.success'),
+                    $resource_id
+                )
+            );
+            return true;
+        }
+        $message = sprintf($this->trans('commands.rest.disable.messages.already-disabled'), $resource_id);
```
edutrul: This is a new string to translate, I am adding this because if you try to disable a rest which is already disabled then a message like 
Rest ID "%s" was already disabled

I think makes sense, 
else
since it is a new string to translate(and takes effort to do it in != languages) then I think I can remove that line and move lines 106-111 into 114 (which is out of "if" conditional) so that it displays always success message without taking care if it is already disabled
What do you think @enzolutions  ?

enzolutions:
Thanks @edutrul

Issue #2912, opened 2016-11-10
[rest:enable] Does not enable as expected


edutrul:
# Problem/Motivation

I was trying to enable a REST, but when I wanted to check it out in /admin/config/services/rest it doesn't enable as expected, it stays in disabled state.

# How to reproduce

Steps:
1. Make sure you have restui contrib module install https://www.drupal.org/project/restui
2. Go to /admin/config/services/rest) enable an entity resource like for instance "content" which name is "entity:node"
3. You will see "Content" is disabled
4. Then execute "drupal rest:enable"
5. After that type "entity:node" and choose method and authentication provider e.g "GET and basic_oauth"
6. Go back to /admin/config/services/rest
7. You will see it is not enabled

Other details:
Drupal version: 8.2
Console version: 1.0.0-rc8

# Solution

I am going to be working on it and soon I'll have a PR asap

edutrul:
I did initial attempt
PR https://github.com/hechoendrupal/DrupalConsole/pull/2919
I'll keep working on it!

edutrul:
PR implementation:
https://github.com/hechoendrupal/DrupalConsole/pull/2919

PR languages:
English: https://github.com/hechoendrupal/drupal-console-en/pull/39
Spanish: https://github.com/hechoendrupal/drupal-console-es/pull/11

jmolivas:
PR merged.

amaisano:
This is still a problem for me on Drupal console 1.6.1 and Drupal 8.4.5 (installed with composer).

PR #2919, opened 2016-11-11
[rest:enable] Fix enable rest resource

edutrul:
What I did:

- Did cleanup
- Add serializer format
- Update translation text from "states" to "method" since doesn't make sense to be "states"
for the following reasons

GET, POST, DELETE, PATCH  are NOT states! they are METHODS see: 

- http://www.restapitutorial.com/lessons/httpmethods.html
- https://en.wikipedia.org/wiki/Representational_state_transfer

and also translation in English is:
"REST States to enable for Rest resource:"
which should be "METHODS to enable for Rest resource"
I'll make sure I do PR for sure!!

So the above is the reason why I am changing "states" to "method"

Please this PR is still in PROGRESS so DO NOT merge till I finish it! thanks!

UPDATE!:
Finally fix is already done!!

### DIFF
--- a/config/services/drupal-core/rest.yml
+++ b/config/services/drupal-core/rest.yml
@@ -11,6 +11,6 @@ services:
       - { name: drupal.command }
   rest_enable:
     class: Drupal\Console\Command\Rest\EnableCommand
-    arguments: ['@plugin.manager.rest', '@authentication_collector', '@config.factory']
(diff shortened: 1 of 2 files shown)


edutrul:
issue in https://github.com/hechoendrupal/DrupalConsole/issues/2912

edutrul:
This is ready for review!!

enzolutions:
Thanks @edutrul

PR #2946, opened 2016-11-15
[BUGFIX 2581] Ensure AnnotationRegistry::registerLoader is correctly …

mickaelperrin:
…defined

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\Console;
 
+use Doctrine\Common\Annotations\AnnotationRegistry;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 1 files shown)


src/Application.php
```diff
@@ -125,6 +126,11 @@ private function registerCommands()
             ->get('application.commands.aliases')?:[];
 
         foreach ($consoleCommands as $name) {
+            // Some commands call AnnotationRegistry::reset, we need to ensure that
```
jmolivas: @mickaelperrin why is this inside the commands iteration?

mickaelperrin: As I said previously, while debugging, I discovered that the registration of some commands call `AnnotationRegistry::reset`. If we want to ensure, that the Annotation is correctly set, we must ensure it is correctly defined before the beginning of the registration of a new command.

jmolivas:
@mickaelperrin Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #2950, opened 2016-11-15
Site:new option (template)

marcelovani:
Here is the implementation
to test, run
drupal site:new /tmp/test3 --composer --template=dennisdigital/drupal-project

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "drupal/console",
+    "name": "marcelovani/console",
     "description": "The Drupal CLI. A tool to generate boilerplate code, interact with and debug Drupal.",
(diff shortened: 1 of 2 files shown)


marcelovani:
We might not need site:new anymore
With this PR its possible to create it as a chain https://github.com/hechoendrupal/DrupalConsole/pull/2961

jmolivas:
@marcelovani Thanks for your contribution. But we are closing this PR, since added as chain file and starting on RC-11 chain files could be registered as commands and you can set the project as placeholder.
https://github.com/hechoendrupal/drupal-console-core/blob/master/config/dist/chain/site-new.yml

This mean of you update to latest version and execute the `init` command the `site:new` will be available.

PR #2988, opened 2016-12-02
Update to Symfony 3.

jmolivas:


composer.json
```diff
@@ -40,15 +40,15 @@
         "alchemy/zippy": "0.3.5",
         "composer/installers": "~1.0",
         "drupal/console-core" : "~1.0",
-        "symfony/css-selector": "~2.7|~2.8",
-        "symfony/debug": "~2.6|~2.8",
-        "symfony/dom-crawler": "~2.8|~3.0",
-        "symfony/http-foundation": "~2.8",
+        "symfony/css-selector": "~2.7|~3.1",
+        "symfony/debug": "~2.6|~3.1",
```
pfrenssen: If Drupal Console was working before with 2.8 versions then I would still keep this mentioned. This can only help in resolving a workable set of dependencies for specific projects. You never know they might have a dependency on css-selector ~2.8 somewhere.

So I suggest to simply mention all supported versions, for example:

```
"symfony/css-selector": "~2.7|~2.8|~3.0|~3.1",
```

jmolivas:
@pfrenssen true I will update this dependency

pfrenssen:
Thanks! Make sure to check all of them, the css-selector one was just an example.

PR #2993, opened 2016-12-03
config schema validation and debugging commands

twhiston:
Added two commands for config validation

config:validate - takes a config name and validates the config against the schema

config:validate:debug - takes a path to a config file, and a path to a schema file and validates the config against the schema. This is super useful for developing modules, as you can test your config without uninstalling and re-installing your module multiple times.

I'm not exactly sure how I should add my translation files now, and I dont really see any in the repo any more, and I couldn't find any docs or issues about the correct way to do this. Please can you check and advise.

### DIFF
--- a/src/Command/Config/ValidateDebugCommand.php
+++ b/src/Command/Config/ValidateDebugCommand.php
@@ -1,0 +1,105 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\Console\Command\Config\ValidateDebugCommand.
+ */
+
+namespace Drupal\Console\Command\Config;
+
+use Drupal\Core\Config\Schema\SchemaCheckTrait;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Command\Command;
+use Drupal\Console\Command\Shared\ContainerAwareCommandTrait;
+use Drupal\Console\Style\DrupalStyle;
+use Symfony\Component\Console\Input\InputArgument;
+use Drupal\Core\Config\TypedConfigManagerInterface;
+use Drupal\Core\Serialization\Yaml;
+
+/**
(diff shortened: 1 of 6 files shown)


jmolivas:
@twhiston Awesome thanks for the PR.

Translations should go at this repo => https://github.com/hechoendrupal/drupal-console-en

twhiston:
Thanks very much, translations removed from this PR and a PR has been made on the other repo https://github.com/hechoendrupal/drupal-console-en/pull/45

src/Command/Config/ValidateDebugCommand.php
```diff
@@ -0,0 +1,105 @@
...
+use Symfony\Component\Console\Input\InputArgument;
+use Drupal\Core\Config\TypedConfigManagerInterface;
+use Drupal\Core\Serialization\Yaml;
+
+/**
+ * Class ValidateDebugCommand.
+ *
+ *@package Drupal\Console\Command\Config
+ */
+class ConfigValidateDebugCommand extends Command {
```
enzolutions: The class name is wrong must to be **ValidateDebugCommand**

twhiston:
Instructions on how to test are basically my first post. 
For config:validate put in a config name and it will check what config is in the db against the schema file for that config, this is useful as you can check for both incorrect data and missing schema, which is a cause of a lot of unit tests failing in contrib modules! 

For config:validate:debug put the path to a yml file of config, and a path to the schema that the first file is meant to adhere to, and it will tell you if it does or not. Basically the debug is to test your config against a schema before you install a module, to make it a bit faster to make sure everything works as it should.

Class name fixed, sorry about that!

twhiston:
Some configs and a schema you can use to test the command (especially the debug one as you wont need to install anything)
[Archive.zip](https://github.com/hechoendrupal/DrupalConsole/files/636247/Archive.zip)

jmolivas:
@twhiston Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #2997, opened 2016-12-06
Create x.install file instead of module when using generate:update

colinappnovation:
# Problem/Motivation

Noticed today that when using **drupal generate:update** with no module or install file present; a module file is created rather than *.install.

# Solution

Modification to generate() function within UpdateGenerator.php to create *.install file as apposed to *.module file. As hook_update_N() should be placed in *.install file by best practice.

### DIFF
--- a/src/Generator/UpdateGenerator.php
+++ b/src/Generator/UpdateGenerator.php
@@ -41,7 +41,7 @@ class UpdateGenerator extends Generator
 
         $this->renderFile(
             'module/src/update.php.twig',
-            $module_path .'/'.$module.'.module',
+            $module_path .'/'.$module.'.install',
             $parameters,
             FILE_APPEND
         );


enzolutions:
@colinappnovation why .install instead of .module, could you point any documentation that suggest this change, please.

colinappnovation:
@enzolutions https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Extension!module.api.php/function/hook_update_N/8.2.x

"For each change in a module that requires one or more actions to be performed when updating a site, add a new implementation of hook_update_N() to your mymodule.install file (assuming mymodule is the machine name of your module). Implementations of hook_update_N() are named (module name)_update_(number)."

timmillwood:
+1 to this hook_update_N() functions should always go in the .install file.

jmolivas:
@colinappnovation Thanks for your contribution, your PR was merged. This will be included on the next release.

@timmillwood thanks for your participation on this PR.

PR #3007, opened 2016-12-12
[multisite:new] Use correct  property in example.sites.php search

rpsu:
Use $this->appRoot, $this->root is empty, which breaks
example.sites.php -file test.

This fixes the issue where `example.sites.php` is not found during  `drupal multisite:new --site-uri=local-2.example.com --copy-install -learning -vvvv local-2.example.com`


jmolivas:
@rpsu Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #3008, opened 2016-12-12
[user:password:hash] relax password service typehint

jacktonkin:
Sites may want to replace the default password service (e.g. to use a wrapper around password_hash() and friends).  Typehint the password service as Drupal\Core\Password\PasswordInterface instead of Drupal\Core\Password\PhpassHashedPassword


jmolivas:
@jacktonkin Nice catch and thanks for your contribution, your PR was merged. This will be included on the next release.

PR #3009, opened 2016-12-12
[user:password:hash] fix typo

jacktonkin:
Calling the hash() method on the password string, not the password service causes fatal error.


jmolivas:
@jacktonkin  Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #3012, opened 2016-12-13
[database:dump help] Correct keys for translation lookup

rpsu:
drupal help database:dump refers to wrong key when looking for (english) translations. This PR fixes the issue and help displays the correct text instead of wrong key with no translation.


jmolivas:
@rpsu nice catch and thanks for your contribution, your PR was merged. This will be included on the next release.

PR #3026, opened 2016-12-18
Remove two unused use statements from the template to generate entity.page.inc

pfrenssen:
The `myentity.page.inc` file which is included when generating a content entity contains two unused use statements. This fails coding standards checks.


enzolutions:
Thanks @pfrenssen for your contributions

PR #3027, opened 2016-12-19
[site:install] Allow empty db password during install

rpsu:
Allow user to omit --db-pass -option ie. set an empty password (instead of assuming ‘root’ as password)


rpsu:
This is useful in local development, since not all environments do set a password for root-user.

jmolivas:
@rpsu Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #3038, opened 2016-12-22
generate:plugin:migrate:source

NickWilde1990:
Currently uses sqlBase so perhaps should consider actually naming it ..:source:sqlbase? (see templates/module/src/Plugin/migrate/source/source.php.twig) english strings PR as well


jmolivas:
@NickWilde1990 Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #3039, opened 2016-12-22
generate:plugin:block: Use question strings

NickWilde1990:
There are question strings defined (mostly) but they aren't used, unlike in other plugin generation commands, change interact() to use them.


jmolivas:
@NickWilde1990 Thanks for your contribution, your PR was merged. This will be included on the next release.

Issue #3041, opened 2016-12-24
config:validate and generate:module have a parameter mismatch


cburschka:
1.0.0-rc12 is installed (both core and console).

I'm getting this exception:

```
ArgumentCountError: Too few arguments to function Drupal\Console\Command\Config\ValidateCommand::__construct(), 0 passed and exactly 1 expected in vendor/drupal/console/src/Command/Config/ValidateCommand.php on line 30 #0 [internal function]: Drupal\Console\Command\Config\ValidateCommand->__construct()
#1 vendor/symfony/dependency-injection/ContainerBuilder.php(927): ReflectionClass->newInstanceArgs(Array)
#2 vendor/symfony/dependency-injection/ContainerBuilder.php(475): Symfony\Component\DependencyInjection\ContainerBuilder->createService(Object(Symfony\Component\DependencyInjection\Definition), 'console.config_...')
#3 vendor/drupal/console/src/Application.php(150): Symfony\Component\DependencyInjection\ContainerBuilder->get('console.config_...')
#4 vendor/drupal/console/src/Application.php(39): Drupal\Console\Application->registerCommands()
#5 vendor/symfony/console/Application.php(123): Drupal\Console\Application->doRun(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#6 vendor/drupal/console/bin/drupal.php(69): Symfony\Component\Console\Application->run()
#7 vendor/drupal/console/bin/drupal(4): require('...')
#8 {main}
```

The most likely cause seems to me to be that config:validate declares no arguments in yaml:

    console.config_validate:
      class: Drupal\Console\Command\Config\ValidateCommand
      tags:
        - { name: drupal.command }

ValidateCommand has a constructor that expects an untyped `$name` argument.

```
class ValidateCommand extends Command
{
    use ContainerAwareCommandTrait;
    use SchemaCheckTrait;
    use PrintConfigValidationTrait;

    public function __construct($name)
    {
        parent::__construct($name);
    }
```

(The remaining packages are all at the versions installed by the composer.lock in Drupal 8.2.4)

cburschka:
This seems to be trivially fixable by removing the __construct stub, since it overrides the parent where $name is optional.

generate:module has a similar bug, in that it asks for an 8th argument named simply $twigTemplate, which is not declared in generate.yml where the service only has seven arguments.

jmolivas:
[...]

PR #3044, opened 2016-12-25
Fix constructor args in two commands (#3041)

cburschka:
- config:validate overrides the parent constructor's optional
  name argument with a constructor that requires it.

- generate:module requires an argument that is not defined in
  the corresponding services file, and should be optional.


jmolivas:

PR #3045, opened 2016-12-27
GenerateModule: PHP7.1 fatal error unused arg

NickWilde1990:
With PHP 7.1 an unused argument without a default value is a fatal error. The twigtemplate argument was introduced but not used in the this context.


NickWilde1990:
(o be exact, ArgumentCountError due to non enough arguments being provided to a function.

jmolivas:
@NickWilde1990 thanks for the PR. Issue fixed with https://github.com/hechoendrupal/drupal-console/pull/3044

PR #3061, opened 2017-01-02
[shell] Relocate command.

jmolivas:


tdm4:
@jmolivas  bin/php-parse and bin/psysh are broken symlinks and should be removed as they are causing problems (See #3186 )

PR #3069, opened 2017-01-03
Allow usage of features:import with features_ui disabled

jorissteyn:
Before this commit, the features:debug command was available without the
UI module enabled but features:import was not. Since the console
commands do not depend on the UI this commit allows feature imports with
the UI module disabled.

@miguel303 - any thoughts or objections?


jmolivas:

PR #3070, opened 2017-01-03
Don't crash on updates in installation profiles

jorissteyn:
When running 'update:execute' on an installation profile, drupal console
crashes because it assumes every update belongs to a module while in
practice installation profiles too can contain updates.


jmolivas:

PR #3078, opened 2017-01-05
Allow commands to return an exit code

jorissteyn:
It is important for any program to exit with a non-zero exit code if it
encountered an error.

In Symfony console commands it's common to return with an exit code in
the execute() method of a command. Drupal Console disregards this exit
code in Application::doRun(), making it harder to use Drupal Console
in a scriptable way.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -53,12 +53,16 @@ class Application extends BaseApplication
         if ($clear === true || $clear === 'true') {
             $output->write(sprintf("\033\143"));
         }
-        parent::doRun($input, $output);
+
+        $exitCode = parent::doRun($input, $output);
+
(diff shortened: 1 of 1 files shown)


jorissteyn:
It's worth noting some existing commands already use 'return 1' for error conditions. With this change Drupal Console will no longer exit with 0 when those commands fail. That might bite some users if they somehow rely on Drupal Console ignoring errors. But most commands don't return with a numeric value,
so those commands will keep failing with exit code 0.

This change allows custom commands to specify an exit code, another improvement would be to change existing commands to return an exit code on error conditions.

jmolivas:
@jorissteyn awesome. thanks for the PR.

jmolivas:

Issue #3083, opened 2017-01-09
[config:import] set exit status on failure.


cburschka:
### Problem/Motivation

For non-interactive usage, it would be useful to allow scripts to react to the exit status of a drupal-console command.

For example if `drupal config:import` fails for any reason, the command will output an error message but otherwise give no indication to the caller that the import failed and (for example) a setup script should be aborted.

### Solution

This might be solved by defining a new exception type that is thrown by a failing command (after showing the appropriate UI errors), then catching this exception from the top-level `$application->run()` call and using `exit(1)` to set the exit status.

cburschka:
Update: Actually, the Symfony console already has exit codes - it's the return value of any function overriding `Symfony\Component\Console\Command::execute()`. So the only fix needed would be to check that all core commands that have their own error handling return 1 if they fail.

Back to the config:import example, this is related to an apparent bug in which the line `$io->success($this->trans('commands.config.import.messages.imported'));` is never executed, because `configImport()` doesn't return anything (neither on success nor failure).

cburschka:
(I see that the support for this was added very recently in #3078, which is great!)

jmolivas:
Closing this one. Since PR #3078 merged.

PR #3084, opened 2017-01-09
[config:import] return exit code on error (#3083)

cburschka:


src/Command/Config/ImportCommand.php
```diff
@@ -127,8 +130,9 @@ private function configImport(DrupalStyle $io, StorageComparer $storage_comparer
             $io->success($this->trans('commands.config.import.messages.already-imported'));
         } else {
             try {
-                $config_importer->import();
                 $io->info($this->trans('commands.config.import.messages.importing'));
+                $config_importer->import();
+                return TRUE;
```
jmolivas: should this be `return 0;` ?

cburschka-pwc: The `execute()` function is expected to return an exit code (1 on failure, 0 or nothing on success), while the private `configImport()` function is expected to return a success boolean (TRUE on success, FALSE or nothing on failure). So it's weird, but consistent with the current usage...

jmolivas: My bad, this is a private scope function. not the execute function.

jmolivas:

PR #3118, opened 2017-01-22
[config:delete] Use StorageInterface.

jmolivas:


kmajzlik:
Thank you for great work!

PR #3128, opened 2017-01-24
[update:execute] Fix a missing variable argument.

### LINKED ISSUE #3108: [update:execute] ignores <module> argument
There's some broken logic in the Command\Update\ExecuteCommand - the module argument is required, but the command always prints the error that no updates exist for the given module, and then runs all updates for all modules anyway.

As the execute() function seems to show, there isn't actually any code that would filter the updates by module:

        $this->module = $input->getArgument('module');
        $updates = update_get_update_list();
        $this->checkUpdates($io);

And in checkUpdates we have...

    private function checkUpdates(DrupalStyle $io)
    {
        if ($this->module != 'all') {
            if (!isset($updates[$this->module])) {

Looks like an improper refactoring: The local $updates variable wasn't turned into an argument or property. Also, even though checkUpdates returns with an error, the main execute() function just continues. ;)

(There might be some more errors in this; might check later.)

cburschka-pwc:
Fixes #3108 and allows `drupal update:execute <module>` to work.

### DIFF
--- a/src/Command/Update/ExecuteCommand.php
+++ b/src/Command/Update/ExecuteCommand.php
@@ -127,7 +127,7 @@ class ExecuteCommand extends Command
         drupal_load_updates();
         update_fix_compatibility();
         $updates = update_get_update_list();
-        $updates = $this->checkUpdates($io, $updates);
+        $this->checkUpdates($io, $updates);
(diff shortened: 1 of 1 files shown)


cburschka-pwc:
(Fixing this in a new PR, #3138, as there are multiple closely related bugs that would be impractical to fix independently.)

PR #3134, opened 2017-01-26
Cannot install modules due to improperly checked requirements.

david-garcia-garcia:
After last update module requirements are not properly verified, and most modules fail to install.


jmolivas:
@david-garcia-garcia which modules can be used to test this PR before/after?

david-garcia-garcia:
Core's simpletest module will fail to install without this patch.

jmolivas:
@david-garcia-garcia Thanks for your contribution, your PR was merged. This will be included on the next release.

Issue #3137, opened 2017-01-27
[update:execute] Module updates are applied repeatedly


cburschka-pwc:
If a module has multiple pending schema updates, then each of them will be applied multiple times. 

- `module` has pending updates numbered 8200, 8201, 8202, 8203.

`drupal update:execute module` applies:

- 8200
- 8200, 8201
- 8200, 8201, 8202
- 8200, 8201, 8202, 8203

This is because of bad logic in the update execution function. For each pending update, the code iterates from the initial schema version to the pending function.

You need to either apply each pending update, or select the last pending update and iterate, but not both. The former seems a lot cleaner, especially if the pending updates are non-consecutive for some reason.

cburschka-pwc:
The post updates have the same bug, except it's made hilariously worse by the... creative way PHP handles for loops over string keys, apparently.

```
 Executing update function "create_rest_resource_config_entities" of module "rest"
 Executing update function "create_rest_resource_config_entities" of module "rest"
 Executing update function "create_rest_resource_config_entitiet" of module "rest"
 Executing update function "create_rest_resource_config_entitieu" of module "rest"
 Executing update function "create_rest_resource_config_entitiev" of module "rest"
 Executing update function "create_rest_resource_config_entitiew" of module "rest"
 Executing update function "create_rest_resource_config_entitiex" of module "rest"
 [... repeat forever]
```

(Aside from that, post updates do not respect the `<module>` argument, but that's a separate issue.)

jmolivas:
Fix included on RC-16.

PR #3138, opened 2017-01-28
[update:execute] Errors in update code (#3108, #3137)

cburschka:
This fixes a few bugs (particularly #3108 and #3137) in applying schema updates and post-update functions. As a result, the update command will also return exit code 1 in case of errors.


jmolivas:

PR #3207, opened 2017-02-28
[generate:command] Add required argument to ExtensionManager::getList()

PatrickBauer:
There is a bug inside the ExtensionTrait which is only used by the generate:command command right now. It calls ExtensionManager::getList() without an argument, resulting in an `ArgumentCountError: Too few arguments to function Drupal\Console\Extension\Manager::getList()`.

This pull request adds a parameter to the function call. Alternatively the function could provide a default parameter.

Issue: https://github.com/hechoendrupal/drupal-console/issues/3209


jmolivas:
@PatrickBauer Thanks for your contribution, your PR was merged. This will be included on the next release.

Issue #3229, opened 2017-03-15
[template] Use short array syntax in all templates


s-robertson:
### Problem/Motivation
Currently the templates in Drupal Console use a combination of the short and long array syntax. Sometimes, different syntaxes are even used in the same file. For consistency, one syntax should be used across all templates. 

Short syntax makes the most sense as the Drupal core contributors have already decided to switch to short syntax for their coding standards and, as of 8.3, all of core's code has been switched to the short syntax: https://www.drupal.org/node/2776975

### Solution
There doesn't appear to be too many places in the code where the long array syntax is being used, so it should be fairly straightforward to convert those.

PR #3230, opened 2017-03-15
[template] Use short array syntax in all templates

s-robertson:
Pull request for #3229 

### Problem/Motivation
Currently the templates in Drupal Console use a combination of the short and long array syntax. Sometimes, different syntaxes are even used in the same file. For consistency, one syntax should be used across all templates. 

Short syntax makes the most sense as the Drupal core contributors have already decided to switch to short syntax for their coding standards and, as of 8.3, all of core's code has been switched to the short syntax: https://www.drupal.org/node/2776975


jmolivas:
@s-robertson Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #3231, opened 2017-03-16
[site:install] Register missing service

ruloweb:
The http_client service is missing, so the site:install command doen't work


jmolivas:
@ruloweb Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #3235, opened 2017-03-19
Remove duplicated getType() from generated content entities

pfrenssen:
Currently when a content entity is generated that supports bundles, this piece of code is generated in it:

```
/**
 * {@inheritdoc}
 */
public function getType() {
  return $this->bundle();
}
```

This getter is duplicating `EntityInterface::bundle()` which is the standard way of getting the bundle from any entity.

Having this `getType()` available only on entities generated by Drupal Console is confusing, since it is missing from core entities. It also doesn't serve any purpose, it is just duplicating existing functionality.


jmolivas:
@pfrenssen nice catch. Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #3236, opened 2017-03-19
Leverage methods and base field definitions from RevisionLogEntityTrait

pfrenssen:
Currently when generating revisionable content entities we are duplicating a number of methods from `RevisionableContentEntityBase`. This class provides the `RevisionLogEntityTrait` which already contains the following methods so we do not need to duplicate them:

* `getRevisionCreationTime()`
* `setRevisionCreationTime()`
* `getRevisionUser()`
* `setRevisionUser()`
* `getRevisionUserId()`
* `setRevisionUserId()`

Additionally, it already defines base field definitions regarding which are duplicated by us but named differently. This causes duplicate field definitions to be used for these entities:

* `revision_created` (duplicated as `revision_timestamp`)
* `revision_user` (duplicated as `revision_uid`)

It looks like our current implementation is based on that of the `Node` entity which predates the `RevisionLogEntityTrait` which was added to 8.2.x. Even today the `Node` entity is not extending `RevisionableContentEntityBase` because of backwards compatibility reasons.

### DIFF
--- a/templates/module/src/Controller/entity-controller.php.twig
+++ b/templates/module/src/Controller/entity-controller.php.twig
@@ -97,7 +97,7 @@ class {{ entity_class }}Controller extends ControllerBase implements ContainerIn
         ];
 
         // Use revision link to link to revisions that are not active.
-        $date = \Drupal::service('date.formatter')->format($revision->revision_timestamp->value, 'short');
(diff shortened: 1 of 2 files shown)


jmolivas:
@pfrenssen awesome the less code we have to maintain the better.

jmolivas:

PR #3244, opened 2017-03-29
Improve docblocks for generated constructors

pfrenssen:
When generating classes the documentation of the constructors is often missing or not according to documentation standards. The most typical phrase that is used in core is:

```
/**
 * Constructs an ImageFormatter object.
 */
```

but since we have to use either 'a' or 'an' depending on whether the class name starts with a vowel, I have opted for the following wording to avoid complexity:

```
/**
 * Constructs a new ImageFormatter object.
 */
```

This will be grammatically correct regardless of the naming of the class.


jmolivas:

PR #3246, opened 2017-03-29
Use short array syntax for generated options

pfrenssen:
When generating a form the resulting code is formatted with the short array syntax everywhere, but the options for a `select` form field are still using the old syntax:

```
    $form['category'] = [
      '#type' => 'select',
      '#title' => $this->t('Category'),
      '#options' => array('Technical problem / bug report' => $this->t('Technical problem / bug report')),
      '#size' => 5,
    ];
```

This PR ensures that short array syntax is also used for the select options.


jmolivas:
@pfrenssen nice catch. Another PR fixes templates but this was left behind.

jmolivas:

PR #3286, opened 2017-04-28
[console] Standardize the second argument of Command::addOption().

### LINKED ISSUE #3174: [console] review the usage of Command::addOption()
Hi,

when working on #3092 I noticed that in some cases `false` is passed as the second argument of `Command::addOption()`, and sometimes an empty string is passed.

This made me wonder that some reader might naively end up thinking that this second argument could be the default value of the option in case the option mode is `InputOption::VALUE_OPTIONAL` (for a moment I did), but that would be wrong.

The second argument is a shortcut name for the command, and the documentation says that it can be `null` if there is none:
http://api.symfony.com/3.1/Symfony/Component/Console/Command/Command.html#method_addOption
A possible default value goes in the fifth argument if needed.

So what about using `null` everywhere for the shortcut argument of `addOption()`?
Just an idea.

ao2:
Fix #3174.

The second argument of Symfony\Component\Console\Command\Command::addOption()
is the shortcut name of the option, and it can be set to null when not used,
as reported in the documentation:
http://api.symfony.com/3.1/Symfony/Component/Console/Command/Command.html#method_addOption

However, currently sometimes false or an empty string are used when the option
does not have a shortcut name. This could cause some confusion.

Use null everywhere as the second argument when the option does not have
a shortcut name.

The refactoring was automated with sed, and validated with php-parser with these scripts:
https://git.ao2.it/experiments/php-drupal-console-code-refactoring.git/tree

I could not use php-parser alone because it looses the original code formatting (until https://github.com/nikic/PHP-Parser/issues/344 gets fixed), but I still used it to validate what I did with sed because I didn't trust myself with sed and regexes.

Ciao ciao,
   Antonio

### DIFF
--- a/src/Command/Config/ExportCommand.php
+++ b/src/Command/Config/ExportCommand.php
@@ -63,17 +63,17 @@ class ExportCommand extends Command
             )
             ->addOption(
                 'tar',
-                false,
+                null,
                 InputOption::VALUE_NONE,
                 $this->trans('commands.config.export.arguments.tar')
             )->addOption(
                 'remove-uuid',
-                '',
+                null,
                 InputOption::VALUE_NONE,
                 $this->trans('commands.config.export.single.options.remove-uuid')
             )->addOption(
                 'remove-config-hash',
-                '',
+                null,
                 InputOption::VALUE_NONE,
                 $this->trans('commands.config.export.single.options.remove-config-hash')
             );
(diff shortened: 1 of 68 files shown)


jmolivas:
@ao2 Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #3295, opened 2017-05-04
Print launcher version, addresses hechoendrupal/drupal-console#3164

nvaken:
With this commit, drupal console will attempt to fetch and use a $launcherVersion variable. This commit needs another commit within drupal-console-launcher in which I add this variable to the global scope. 

After this, drupal console will be able to print the launcher's version, which looks like:
image

This adresses issue: hechoendrupal/drupal-console#3164

### DIFF
--- a/bin/drupal.php
+++ b/bin/drupal.php
@@ -63,6 +63,9 @@ if (!$container) {
     exit(1);
 }
 
-$application = new Application($container);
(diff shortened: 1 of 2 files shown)


jmolivas:
@nvaken Thanks for your contribution, your PR was merged. This will be included on the next release.

PR #3311, opened 2017-05-14
[generate:entity:content] Add 'view_builder' to config entity annotation template. Fix #3129.

### LINKED ISSUE #3129: [generate:entity:content] Missing view_builder handler for custom entity with bundles
### Problem
When creating a custom entity with bundles support, you are not able to access entities through REST api.

### How to reproduce
- Drupal version: 8.2.5.
- Console version: 1.0.0-rc14
- Console Launcher version (using [drupal-composer](https://github.com/drupal-composer/drupal-project) project).

**Steps to reproduce**

- Create a custom content entity `Music` using `generate:entity:content` command with bundles support enabled. It will create two entity classes `Music` and `MusicType`
- Create a bundle `Rock`. Add some content for Rock bundle.
- Enable REST resource for `Music`in configuration which will be something like `/admin/structure/music/{music}`.
- Make a GET request with REST api like `/admin/structure/music/1`. 
- You will get error something like `{"message":"A fatal error occurred: Route entity.music_type.canonical does not exist."}`

### Solution
As mentioned in [this](https://www.drupal.org/node/2826230) issue, this is happening because `Drupal\Core\Entity\Routing\DefaultHtmlRouteProvider::getCanonicalRoute()`requires `view_builder` to create a canonical url, which is missing from `MusicType` class. 
If you manually add `"view_builder" = "Drupal\Core\Entity\EntityViewBuilder"` in handlers list of `MusicType` class, it works fine.

jall:
This is a patch for #3129.

This allows content entities with bundles to be exposed through the core REST API without throwing errors of the form: 
```
Route entity.{{entityBundleType}}.canonical does not exist.
```

Steps to reproduce & fix manually have been well described by @tahirm in the linked issue, so I won't re-explain them here.

This patch adds a default 'view_builder' to all config entity definitions. It uses the Core `Drupal\Core\Entity\EntityViewBuilder` & doesn't bother defining a custom extension class in the same way as, for example, the 'list_builder': `Drupal\{{ module }}\{{ entity_class }}ListBuilder` as I haven't come across a use case for this level of flexibility yet. 

I checked Core to see where it extends `EntityViewBuilder` rather than using it directly and found of the >30 Drupal Core config entities, only 2 have 'view_builder' defined:
* `\Drupal\tour\Entity\Tour` => `Drupal\tour\TourViewBuilder`
* `\Drupal\block\Entity\Block` => `Drupal\block\BlockViewBuilder`

These both appear to be bundle config entities, which fits with the use case being fixed here.

I was concerned that adding 'view_builder' unconditionally to `entity.php.twig` might be too far reaching, as it doesn't look like a config entity needs it defined unless they specify a canonical route. However, it appears that the only usage of this template is for adding content entity's bundle classes in `src/Generate/EntityConfigGenerator`, so I believe this patch will only affect what it should.

### DIFF
--- a/templates/module/src/Entity/entity.php.twig
+++ b/templates/module/src/Entity/entity.php.twig
@@ -24,6 +24,7 @@ use Drupal\Core\Config\Entity\ConfigEntityBase;
  *   id = "{{ entity_name }}",
  *   label = @Translation("{{ label }}"),
  *   handlers = {
+ *     "view_builder" = "Drupal\Core\Entity\EntityViewBuilder",
  *     "list_builder" = "Drupal\{{ module }}\{{ entity_class }}ListBuilder",
  *     "form" = {
  *       "add" = "Drupal\{{ module }}\Form\{{ entity_class }}Form",


jmolivas:
@jall Thanks for your contribution, your PR was merged. This will be included on the next release.

davereid:
I think this should be reverted. This line is not necessary on the majority of config entities.

jall:
@davereid I can see why a `view_builder` would be dead code for many config entities which are not exposed/rendered, but does it introduce any performance impact or have other negative impacts other than cleanliness?

I raised this PR because I was generating a lot of bundle-able entities and exposing them as REST resources, & it was quite irritating having them fail by default before this patch.

I know very little about the bigger picture entity system but a lot of the other generated entity code from Drupal console seems be fairly verbose boilerplate & possibly unnecessary for many use cases, so I didn't think this would be any different.

jmolivas:
We can probably pass some flags to decide which code to generate.

PR #3341, opened 2017-06-09
Use fully qualified namespace for @use statements

pfrenssen:
`@use` statements are part of documentation and are shown in rendered documentation. They should use fully qualified namespaces. This also ensures that the developer can navigate to the classes in their IDE.


jmolivas:

PR #3344, opened 2017-06-12
Collection routes are provided by core

pfrenssen:
When generating an entity we were providing the collection route in the `HtmlRouteProvider` but this is no longer necessary since core now provides the route. This was added in Drupal 8.2.0 in [Issue #2767853 by tstoeckler: Provide a route for collections of entities](https://www.drupal.org/node/2767853).

### DIFF
--- a/templates/module/src/entity-route-provider.php.twig
+++ b/templates/module/src/entity-route-provider.php.twig
@@ -29,39 +29,8 @@ class {{ entity_class }}HtmlRouteProvider extends AdminHtmlRouteProvider {% endb
   public function getRoutes(EntityTypeInterface $entity_type) {
     $collection = parent::getRoutes($entity_type);
 
-    $entity_type_id = $entity_type->id();
-
-    if ($collection_route = $this->getCollectionRoute($entity_type)) {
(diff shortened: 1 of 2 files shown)


templates/module/src/entity-route-provider.php.twig
```diff
@@ -29,39 +29,8 @@ class {{ entity_class }}HtmlRouteProvider extends AdminHtmlRouteProvider {% endb
   public function getRoutes(EntityTypeInterface $entity_type) {
     $collection = parent::getRoutes($entity_type);
 
-    $entity_type_id = $entity_type->id();
-
-    if ($collection_route = $this->getCollectionRoute($entity_type)) {
-      $collection->add("entity.{$entity_type_id}.collection", $collection_route);
-    }
+    // Provide your custom entity routes here.
```
pfrenssen: For config entities the `HtmlRouteProvider` has now become unnecessary. I have replaced the logic with this comment that explains that custom routes can be added here.

Shall we keep this dev hint or would it be better to just remove the file?

pfrenssen:
The test failed because it was cancelled for some reason. When it is restarted it will probably turn green again :)

jmolivas:

PR #3350, opened 2017-06-13
Remove @package declarations from generated code

pfrenssen:
Many of the classes are generated with an `@package` declaration. This is a standard from PhpDoc and is adopted by other popular CMS's like Wordpress but not by Drupal. It is not defined in the Drupal coding standards, and is not used by Drupal core and contrib modules. This information duplicates the namespace declaration, and is completely useless.

This probably found its way into Drupal Console because it is added to class docblocks automatically by PHPStorm.


jmolivas:

PR #3351, opened 2017-06-14
Quoting route names is not necessary

pfrenssen:
This is a really minor fix. When specifying routes in YAML files it is not necessary to quote them, since they are resolved to route names, and not raw strings. In almost all of our generated code this is handled correctly, save for two instances when generating action links.


jmolivas:

PR #3367, opened 2017-06-30
[console] Load commands from profiles when drupal and profile are not installed.

jmolivas:


jmolivas:
Waiting on this d.o. issue => https://www.drupal.org/node/2889965

PR #3371, opened 2017-07-03
[generate:entity:content] Improve command options

derhasi:
In the current `drupal generate:entity:content` command there are some problems, so you could not generate an entity without any manual confirmation step. The base-path set via `--base-path="..."` currently only is set as default for the question displayed after running the command. With the options `has-bundles`, `is-translatable` and `revisionable` there is the problem of not beeing able to disable those settings via the command line.

With the given PR I try to fix those problems, so a command like below, will run without the need of any manual confirmations:
```
drupal generate:entity:content --module=kswertung --entity-class=ResultEntity --entity-name=ksresult --label="Result" --base-path="/kswertung" --revisionable --has-bundles=0 --is-translatable=false
```

The PR adds the possibility to disable an option by setting it to _0_ or _false_.

### DIFF
--- a/src/Command/Generate/EntityCommand.php
+++ b/src/Command/Generate/EntityCommand.php
@@ -150,12 +150,12 @@ abstract class EntityCommand extends Command
         // --base-path option
         $base_path = $input->getOption('base-path');
         if (!$base_path) {
-            $base_path = $this->getDefaultBasePath();
+            $base_path = $io->ask(
+                $this->trans('commands.'.$commandKey.'.questions.base-path'),
+                $this->getDefaultBasePath()
+            );
         }
-        $base_path = $io->ask(
-            $this->trans('commands.'.$commandKey.'.questions.base-path'),
-            $base_path
-        );
+
         if (substr($base_path, 0, 1) !== '/') {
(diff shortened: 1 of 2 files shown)


jmolivas:
@derhasi thanks for the PR.

By setting an option as `VALUE_NONE` if not passed value is false. 

Not sure if we should ask for those options on the interactive mode, can we just used them as false.

derhasi:
@jmolivas I think the option always should be asked for interactively, when it was not explicitly set in the command line.

But as you cannot pass a value to an option that is set to `VALUE_NONE`, I chose to use `VALUE_OPTIONAL`. With `hasOption()` you can check if the option is used in the command line and otherwise ask for a value interactively.

derhasi:
The implementation of this PR would result in ...
* `drupal generate:entity:content` => ❓ interactive mode
* `drupal generate:entity:content --revisionable` => ✅ revision enabled
* `drupal generate:entity:content --revisionable=0` => ❌ revision disabled
* `drupal generate:entity:content --revisionable=false` => ✅  revision disabled
* `drupal generate:entity:content --revisionable=Baum` => ❌ revision enabled

PR #3372, opened 2017-07-03
[generate:entity:content] Fix revision revert path for untranslatable content

derhasi:
When creating a content entity with revisions enabled but translation disabled, the revision_revert route is not correctly registered.


jmolivas:
@derhasi Thanks for your contribution, your PR was merged. This will be included on the next release.

Issue #3373, opened 2017-07-03
config:diff seemingly cannot be run non-interactively


cburschka-pwc:
```
$ drupal config:diff

 Config directory: [sync]:
  [0] sync
  [1] dev
```

```
$ drupal config:diff -n

                                     
  [RuntimeException]                 
  Directory name must not be empty.  
```

There *is* a default, you can just hit enter on the above prompt; why doesn't it use that in non-interactive mode?

```
$ drupal config:diff sync                                 
                                                                                       
  [UnexpectedValueException]                                                           
  DirectoryIterator::__construct(sync): failed to open dir: No such file or directory  
                                                                                       
```

I have no idea how to use this command now...

cburschka-pwc:
There seem to be two separate bugs here - the first is that the interactive mode sets a default which is not used in non-interactive mode (`CONFIG_SYNC_DIRECTORY`), and the second is that the interactive mode prompts for a `$config_directories` key, but the CLI argument itself is expected to be a path.

My first approach would be adding something like this to `::execute`:

```
        global $config_directories;
        $directory = $input->getArgument('directory') ?: CONFIG_SYNC_DIRECTORY;
        if (array_key_exists($directory, $config_directories)) {
            $directory = $config_directories[$directory];
        }
```

Which allows passing either a key or a path, and defaults to the CONFIG_SYNC_DIRECTORY constant if nothing is entered. But it's somewhat dirty to allow two kinds of value in the same argument (and might have weird effects if a file path happens to be a key in that array), and maybe it would be better to have two different named arguments.

jmolivas:
Fixed with PR https://github.com/hechoendrupal/drupal-console/pull/3388

I tested with:
```
drupal config:diff sync --no-interaction
drupal config:diff ../config/sync --no-interaction
```

jmolivas:
Closing this one. Feel free to re-open if still happening after updating to the latest version.

PR #3388, opened 2017-07-04
[config:diff]:fix bug with directory parameter. (#3373)

GDrupal:
This fixes a bug when passing the directory name argument. Now you can pass rather name / path and it will work.


jmolivas:
@GDrupal Thanks for your contribution, your PR was merged. This will be included on the next release.

I tested by executing:
```
drupal config:diff sync --no-interaction
drupal config:diff ../config/sync --no-interaction
```

PR #3407, opened 2017-07-06
Allow removing modules installed by profile.

### LINKED ISSUE #1916: [module:uninstall] Can't uninstall module if it is in the profile list
Hi, 

 I have a custom profile in my Drupal installation. This profile contains list of modules (https://github.com/DmitryDrozdik/dmaps/blob/master/d8/profiles/pp/pp.info.yml), which should be installed during installation Drupal profile. 

After successful installation, I can not uninstall module(dmaps), added to the profile list. But I should be able to do this. Uninstalling module works well on the Drupal "uninstall page".

cburschka:
Fixes #1916

### DIFF
--- a/src/Command/Module/UninstallCommand.php
+++ b/src/Command/Module/UninstallCommand.php
@@ -172,6 +172,7 @@ class UninstallCommand extends Command
         }
 
         if (!$force = $input->getOption('force')) {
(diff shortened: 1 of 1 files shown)


jmolivas:

PR #3408, opened 2017-07-06
Allow inherited ModuleInstaller classes to InstallDependencyCommand.

dpagini:
I'm getting a Drupal Console error in my project b/c I'm extending the Drupal "ModuleInstaller" class. My code is then passing in my overridden class via services to Drupal Console. I'm not 100% sure this is right, but this feels like it should be accepting the ModuleInstallerInterface instead of just ModuleInstaller. I'm not that familiar with "ProxyClass"es which is what I'm removing, so I'm not confident this is definitely correct. Please advise if you don't think this is an appropriate bugfix.

### Other:
* Fixed comment styles.
* Added `@param` to constructor comment.


jmolivas:
Exactly we should be injecting interfaces instead of concrete classes.

jmolivas:
@dpagini Thanks for your contribution, your PR was merged. This will be included on the next release.

Issue #3438, opened 2017-07-15
Space in `__con struct` in skeleton.php.twig


clrockwell79:
There is a typo - an extra space - in the template within skeleton.php.twig

`template` 

### Problem/Motivation
Using `drupal generate:plugin:skeleton` results in an extra space in the plugin's `__construct` method, which results in an error.

### How to reproduce
Use `drupal generate:plugin:skeleton`, e.g. `drupal generate:plugin:skeleton --module="commerce_form_customizations" --plugin-id="commerce_checkout_pane" --class="CustomLoginRegisterPane"`

**Details to include:**
- Drupal version 8.3.2
- Console version I'm using 1.0.0-rc19 but it's also in master and 1.0.0-rc24.
- Console Launcher version 1.0.0-rc23.

### Solution
Submitting a pull request

jmolivas:
Fixed with https://github.com/hechoendrupal/drupal-console/pull/3439

PR #3439, opened 2017-07-15
Fix space in `__con struct`

clrockwell79:
For #3438


clrockwell79:
I'm not sure if I botched something here - I didn't mean for b5f4210 to be there

PR #3445, opened 2017-07-17
Add cache:tag:invalidate command.

cburschka:
This is a command for accessing Drupal's cache tag invalidator, which allows clearing caches more specifically than complete deletion, if you know what tags a cache entry is dependent on.

For example, when you're debugging a view, you can invalidate all related cache entries with `drupal cache:tag:invalidate config:views.view.frontpage` without having to actually trigger the view's save operation.


jmolivas:

Issue #3365, opened 2017-06-28
Upgrading to rc23 requires downgrading doctrine/collections


cburschka-pwc:
In #3272 (rc17), the version requirement for doctrine/collections was relaxed in order to allow using doctrine/collections 1.4.0. Then in #3362 (rc23), this was reverted to force 1.3.x again.

Given that between rc17-rc22, Drupal Console has been working fine with 1.4, I'm somewhat confused that in rc23 it was apparently necessary to downgrade back, even though 1.4 should be fully backward-compatible.

Is there a specific bug that required this? The two packages seem to work fine when installed without composer, but I haven't tested it much.

```
$ composer require drupal/console ^1.0.0-rc23 drupal/console-core ^1.0.0-rc23 drupal/console-en ^1.0.0-rc23
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for drupal/console ^1.0.0-rc23 -> satisfiable by drupal/console[1.0.0-rc23].
    - Conclusion: remove doctrine/collections v1.4.0
    - Conclusion: don't install doctrine/collections v1.4.0
    - drupal/console 1.0.0-rc23 requires doctrine/collections 1.3.* -> satisfiable by doctrine/collections[1.3.x-dev, v1.3.0].
    - Can only install one of: doctrine/collections[1.3.x-dev, v1.4.0].
    - Can only install one of: doctrine/collections[v1.3.0, v1.4.0].
    - Installation request for doctrine/collections (locked at v1.4.0) -> satisfiable by doctrine/collections[v1.4.0].
```

jmolivas:
@cburschka we detected `1.4.0` requires php `^5.6 || ^7.0`

1.4.0
https://github.com/doctrine/collections/blob/v1.4.0/composer.json#L16

1.3.0
https://github.com/doctrine/collections/blob/v1.3.0/composer.json#L16

Since latest stable release if Drupal 8 still requires `>=5.5.9` 
https://github.com/drupal/drupal/blob/8.3.4/core/composer.json#L7

We decided to switch back

cburschka-pwc:
[...]

PR #3447, opened 2017-07-21
Update composer package type

gapple:
Since Drupal Console is now installed as a project requirement instead of as a standalone application, my understanding is that the `type` property in `composer.json` should be updated.

- `composer.lock` is ignored if a package is installed within a project as a dependency.
- The [Packagist page](https://packagist.org/packages/drupal/console) displays the setup command for Drupal Console as `composer create-project drupal/console` instead of `composer require drupal/console`.


cburschka:
If composer.lock is removed, this could mean DC can safely require `doctrine/collections` as ^1.3 rather than 1.3.0 again, leaving it to the local composer to resolve to the versions that are compatible with the installed PHP version.

(#3365)

jmolivas:
@gapple Thanks for your contribution, your PR was merged. This will be included on the next release.

jmolivas:
@gapple @cburschka Should we send another PR to remove the `composer.lock` file to fix the `doctrine/collections` issue?

PR #3451, opened 2017-07-24
[console] Fix annotation does not exists.

jmolivas:


kgaut:
Can confirm that this patch fixed my issue 
```

PR #3454, opened 2017-07-25
Add autoload path for when console is symlinked in vendor

gapple:
When trying to test making corresponding changes to multiple components of Drupal Console, I used [drupal-project](https://github.com/drupal-composer/drupal-project), and added some repository overrides in composer.json to use the local packages:

```
{
    "repositories": [
        {
            "type": "path",
            "url": "/Users/geoff/git/drupal-console"
        },
        {
            "type": "path",
            "url": "/Users/geoff/git/drupal-console-core"
        },
        {
            "type": "path",
            "url": "/Users/geoff/git/drupal-console-extend-plugin"
        }
}
```

However, this means that when searching for the autoloader, `__DIR__` corresponds to the target of the symlinks rather than to the path within the drupal-project's vendor directory and execution fails.

This PR adds another path to check for the autoloader based on the current working directory and the path to the `drupal` script, which is unnaffected by the symlinks.

### DIFF
--- a/bin/drupal.php
+++ b/bin/drupal.php
@@ -20,8 +20,12 @@ if (file_exists(__DIR__ . '/../autoload.local.php')) {
     include_once __DIR__ . '/../autoload.local.php';
 } else {
     $autoloaders = [
+        // When a dependency of another project (vendor/drupal/console/bin/drupal.php)
         __DIR__ . '/../../../autoload.php',
-        __DIR__ . '/../vendor/autoload.php'
+        // When console is the root project (bin/drupal.php)
+        __DIR__ . '/../vendor/autoload.php',
+        // When symlinked in vendors/, __DIR__ is not within the root project.
+        dirname($_SERVER['PWD'] . '/' . $_SERVER['SCRIPT_NAME']) . '/../autoload.php',
     ];
 }
 


jmolivas:
@gapple we add a command and instructions to achieve this without all this changes, can you try this instructions?

https://docs.drupalconsole.com/en/contributing/running-the-project.html

jmolivas:
The project search at `bin/drupal.php` for an `autoload.local.php` file
https://github.com/hechoendrupal/drupal-console/blob/master/bin/drupal.php#L19-L20

The new `develop:create:symlinks` provided by the `drupal/console-develop` project, takes care of copying the provided file at the repo [`autoload.local.php.dist`](https://github.com/hechoendrupal/drupal-console/blob/master/autoload.local.php.dist) as `autoload.local.php`

gapple:
That's what I get for diving in without reading the documentation 😝.

`autoload.local.php` worked as well, though I needed to copy it manually due to my setup.  I'll write up my strategy for using composer repositories and version aliasing to handle the package linking another time.

jmolivas:
@gapple We are devs, we only read the docs when everything else fails ;)

Thank you for the great working with the symfony dependencies it was one of the main blockers to tag the stable release.

PR #3185, opened 2017-02-16
[config:export] Export all config collections (#3182)

cburschka-pwc:
In particular, this includes the language.* collections provided by the config_translation module, and fixes a bug wherein config:export followed by config:import will delete translations.

Issue #3182, opened 2017-02-16
[config:export] misses the config translation entities


cburschka-pwc:
When exporting the configuration via the site, the translated objects and entities are exported as `language/XY/ID.yml` files.

drupal config:export must export these as well - if it doesn't, they will get lost on subsequent imports.

cburschka-pwc:
Some investigation shows that the code I'm adding was actually in the class until removed by 186db04dc7b983b27ad0a4188564ec689a7a9b56 in November '15. The only indication I can find in the commit message is "removing unused code". Was that intentional?

The above change adds the collection export back, and it's definitely required for the config_translation strings to get properly exported.

(I haven't checked config:import yet, only the web import.)

jmolivas:
@cburschka-pwc  PR https://github.com/hechoendrupal/drupal-console/pull/3482 merged. Can we consider this one fixed?

enzolutions:
@cburschka-pwc Can we consider this one fixed?

cburschka:
I think that's a safe assumption, based on the PR being merged, though it's been at least two years since I looked at this code. ;-)

PR #3482, opened 2017-08-10
[config:export] Create paths when exporting collections.

cburschka:
When I added the code for exporting config translations in #3182 / #3185, I forgot some `mkdir` calls. ;) So translations are only exported if the language folders already exist, but not when you're exporting for the first time or after creating a new language. (Tar mode is unaffected)


jmolivas:
@cburschka Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3483, opened 2017-08-10
Fall back to environment editor if not set in console config

### LINKED ISSUE #1401: [config:edit] Respect the $EDITOR environment variable
I'm getting vim even though $EDITOR is set to nano. Even removing the line from the config file doesn't help, because it's hard-coded via `$config->get('application.editor', 'vi');`. That's probably kinda silly because it means the part of the code that's supposed to use the environment variable is unreachable.

``` php
        $editor = $config->get('application.editor', 'vi');

        if ($editor != '') {
            return trim($editor);
        }
       // this will never be reached:
        $processBuilder = new ProcessBuilder(array('bash'));
        $process = $processBuilder->getProcess();
        $process->setCommandLine('echo ${EDITOR:-${VISUAL:-vi}}');
```

gapple:
Fixes #1401

### DIFF
--- a/src/Command/Config/EditCommand.php
+++ b/src/Command/Config/EditCommand.php
@@ -166,7 +166,7 @@ class EditCommand extends Command
     protected function getEditor()
     {
         $config = $this->configurationManager->getConfiguration();
-        $editor = $config->get('application.editor', 'vi');
+        $editor = $config->get('application.editor', '');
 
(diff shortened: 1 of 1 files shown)


jmolivas:
@gapple Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3492, opened 2017-08-21
Don't wipe cache during cron run

### LINKED ISSUE #3480: Cron Wipes Cache
### Problem/Motivation
When running cron, caches are always wiped. This is less will cause unnecessary rebuilds of expensive Drupal internals caches and containers much more often that would be advised.

### How to reproduce
run cron. its in the code.
https://github.com/hechoendrupal/drupal-console/blob/master/src/Command/Cron/ExecuteCommand.php#L124

### Solution
Don't chain cache clear. If someone wants the ill advised behavior of clearing cache with cron run just run 2 commands in your cron.

neclimdul:
Fix #3480

Wiping cache manually can kill site performance so forcing it during an
operation that will run very often on most sites could be very damaging.

If a user _needs_ to clear caches at set intervals they can still do so using
the explicit cache commands.

### DIFF
--- a/config/services/cron.yml
+++ b/config/services/cron.yml
@@ -1,7 +1,7 @@
 services:
   console.cron_execute:
     class: Drupal\Console\Command\Cron\ExecuteCommand
-    arguments: ['@module_handler', '@lock', '@state', '@console.chain_queue']
+    arguments: ['@module_handler', '@lock', '@state']
     tags:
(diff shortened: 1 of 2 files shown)


jmolivas:
@neclimdul Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3507, opened 2017-08-31
Skip the reset password page and just login.

arturolinares:
When creating a one time user login link a reset password page is shown with a "login" button to actually login. This change skips that page and just logins directly into the user account.


jmolivas:
@arturolinares Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3509, opened 2017-09-02
Adds aegir support.

ttadegraaff:
Hi, this code works for me, and I expect it might work for others too. That said, I'm a beginner in object oriented programming and Drupal 8, so there probably is a nicer way to get this issue fixed. The database settings are created in Drupal\Core\Site\Settings, and this is a final class. As I understand it this makes it impossible to extend. I ended up copying code from Drupal core to drupal console, which is not nice.

The database settings are read from drushrc.php which is located in the site directory (same location as settings.php).

This is how the database settings exist in my drusrc.php:

$options['db_type'] = 'mysql';
$options['db_host'] = 'localhost';
$options['db_port'] = '3306';
$options['db_passwd'] = 'averystrongpass';
$options['db_name'] = 'thedbname';
$options['db_user'] = 'thedbusername';

I expect aegir always makes the settings available in this way in drushrc.php, but I'm not sure. I haven't looked into the aegir project to check how drushrc.php is created. So there is a possibility that in other environments the settings are stored differently. Nor do I know if there are situations where aegir sites have more then 1 database connection, and how this works if so.

### DIFF
--- a/src/Bootstrap/ConsoleSettings.php
+++ b/src/Bootstrap/ConsoleSettings.php
@@ -1,0 +1,178 @@
+<?php
+
+namespace Drupal\Console\Bootstrap;
+
+use Drupal\Component\Utility\Crypt;
+use Drupal\Core\Database\Database;
+
+/**
+ * Read only settings that are initialized with the class.
+ *
+ * @ingroup utility
+ */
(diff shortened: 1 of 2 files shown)


bserem:
Shall this be merged?

PR #3518, opened 2017-09-07
[update:execute] Warn and return 0 if there are no pending updates

phenaproxima:
### Problem/Motivation
A recent change to the update:execute command has made me a sad panda. As of 1.0.2, if there are no pending updates, an error is output and 1 is returned. This broke our Travis CI builds over at acquia/lightning: https://travis-ci.org/acquia/lightning/jobs/272902792

### Solution
I discussed this problem briefly with @jmolivas on Slack and he said it should warn, not error, and return 0.


jmolivas:
@phenaproxima Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3525, opened 2017-09-18
Allow dependencies of parent profiles to be uninstalled

balsama:
Drupal core #1356276 allows for profiles to be set as a parent profile and for child profiles to inherit from parents. This allows for more than one profile to be installed on a site.

Drupal (and Drush and Drupal Console) has always made exceptions for dependencies of Profiles in that you can actually uninstall dependencies of Profiles without uninstalling the Profile itself. This patch makes the same exception for parent profiles. Or rather, adds all profiles to a list of dependents to exclude instead of just the active/last profile.

It includes a check to see if the patch has been applied to core and works around it if not.

### DIFF
--- a/src/Command/Module/UninstallCommand.php
+++ b/src/Command/Module/UninstallCommand.php
@@ -170,11 +170,22 @@ class UninstallCommand extends Command
         }
 
         if (!$force = $input->getOption('force')) {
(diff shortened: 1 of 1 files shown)


jmolivas:
@balsama Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3530, opened 2017-09-22
[generate:entity:content] Omit 'access /entity type/ overview' permission since it does nothing.

laradevitt:
If I understand things correctly, access to content entity listings/collection used to be provided by the generated entity, but was recently [passed off to core](https://github.com/hechoendrupal/drupal-console/pull/3344). This means that unless the route provider is modified after generation, it's the *administer foo entities* permission that opens the door, not *access foo overview*.

If I've gotten this right - I'm a relative newb to entity building in D8 - it's probably confusing to keep the latter permission lying around.


jmolivas:
@laradevitt Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3546, opened 2017-10-14
Use interfaces for DI.

dpagini:
Similar to #3408. Use interfaces for dependency injection.


jmolivas:
@dpagini Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3547, opened 2017-10-19
Properly record ran post_update functions.

NickWilde1990:
If this isn't done, Drupal UI will continue to report that db updates are pending. 
Note, I considered placing this outside of the foreach calls and recording all post updates at once; however that would cause problems if one failed but not the others.


jmolivas:
Nice catch @NickWilde1990 Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3552, opened 2017-10-23
change-default-command 

rpang77:


jmolivas:
@rpang77  Something is strange with this PR since have too many changed files.

cburschka-pwc:
Is this PR backwards? It looks like it wants to merge the master branch into the topic branch.

jmolivas:
This PR does not apply, default command is no longer set and therefore set to `list`.

PR #3555, opened 2017-10-26
Always extend ContentEntityInterface in interface of new content entity

### LINKED ISSUE #3554: [ generate:entity:content ] Created interface should extend ContentEntityInterface instead of RevisionableInterface
### Problem/Motivation
Generating new revisionable entity you get interface that does not extend all the necessary interfaces above it. Because of that, you lack the knowledge of basic entity functions.

### How to reproduce
Create new revisionable entity and notice how the created interface does not contain functions such as `id()` and `save()` etc. basic functions of entity. 

**Details to include:**
- Drupal version `8.3.7`
- Console version `1.0.2` 
- Console Launcher version `Drupal Console version 1.0.2`
### Steps to reproduce
***Notice***: Here I have not used completely default settings, but it can be reproduced with all being default as well. Only important part is that entity shall be revisionable
```
root@96e01a36ca55:/var/www/html/web# ../vendor/bin/drupal geco       
 // Welcome to the Drupal Content Entity generator
 Enter the module name [layout_test]:
 > 
 Enter the class of your new content entity [DefaultEntity]:
 > 
 Enter the machine name of your new content entity [default_entity]:
 > 
 Enter the label of your new content entity [Default entity]:
 > 
 Enter the base-path for the content entity routes [/admin/structure]:
 > 
 Do you want this (content) entity to have bundles (yes/no) [no]:
 > no
 Is your entity translatable (yes/no) [yes]:
 > no
 Is your entity revisionable (yes/no) [yes]:
 > 
```
Now we can create simple code to observe the actual problem:
image
As you can see, IDE (in this case PhpStorm) does not know functions `id()` and `save()` when entity is declared by the interface, there might be some other missing functions as well, but at least these.
The reason is with the fact that the entity does extend `ContentEntityBase`, which implements `ContentEntityInterface`, but `DefaultEntityInterface` does not extend that same. 
image

### Solution
When changing 
```php
interface DefaultEntityInterface extends RevisionableInterface, RevisionLogInterface, EntityChangedInterface, EntityOwnerInterface {
```
to 
```php
interface DefaultEntityInterface extends ContentEntityInterface, RevisionLogInterface, EntityChangedInterface, EntityOwnerInterface {
```
Then the class diagram shows as we want it to be:
image
And now our code does look flawless in IDE:
image

Dropa:
`ContentEntityInterface` does extend `RevisionableInterface`, but also offers knowledge of basic functions such as `id()`, `save()` etc. 

Fixes #3554

### DIFF
--- a/templates/module/src/Entity/interface-entity-content.php.twig
+++ b/templates/module/src/Entity/interface-entity-content.php.twig
@@ -9,11 +9,9 @@ namespace Drupal\{{module}}\Entity;
 {% endblock %}
 
 {% block use_class %}
+use Drupal\Core\Entity\ContentEntityInterface;
 {% if revisionable %}
 use Drupal\Core\Entity\RevisionLogInterface;
-use Drupal\Core\Entity\RevisionableInterface;
-{% else %}
-use Drupal\Core\Entity\ContentEntityInterface;
 {% endif %}
 use Drupal\Core\Entity\EntityChangedInterface;
 use Drupal\user\EntityOwnerInterface;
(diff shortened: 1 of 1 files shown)


jmolivas:
@Dropa Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3558, opened 2017-10-26
Avoid run of Request handle() and use of container at compiler pass to fix issues 3091 and 3553

citlacom:
Hi @jmolivas - @enzolutions, at Bluespark we have a project based on Drupal 8.3.7 and Drupal Console 1.0.0-beta4 that tried to upgrade to Drupal Console 1.0.2 but this caused a lot of fatal errors in our custom implementations of commands. I researched a bit of the existing issues and found that our error was very similar than https://github.com/hechoendrupal/drupal-console/issues/3091 so tried the workaround detailed on that issue to avoid usage of session but we found another fatal error that I reported at https://github.com/hechoendrupal/drupal-console/issues/3553

I tried to compare what have changed between the two versions to identify why this erros was not experimented in previous Drupal Console version and found that the main reason is the handle of the "/" default request which trigger hooks and globally display blocks access visibility rules among other things.

I reviewed all the Bootstrap process and started to experiment until have a bootstrap which don't caused those fatal exceptions. The main changes are:

- Avoid handle() of the request and only process the preHandle().
- Avoid to process build container on rebuildContainer and boot() to have only the initializeContainer on boot().
- Avoid use of container on compiler pass because that causes PHP warnings and in some cases also errors and just discovered that based on https://github.com/symfony/symfony/issues/22125#issuecomment-288734689 is recommended that container is not used at build phase, that error is exactly what I was experimenting in this project.

I have detailed with comments the reasons of all the changes so you can review, I was not really familiar with Drupal Kernel and Bootstrap process until this issue so I leave it to your expertise to decide what changes to preserve. I have tested multiples core and custom commands that are working properly. I would like to run tests for existing commands but couldn't found in documentation how to run tests in local. Can you guide me?

I hope this helps to guys that have been experimenting this weird problems when upgrading to last stable release and thanks for all your great job on the Drupal Console. I love it!

### DIFF
--- a/src/Bootstrap/AddServicesCompilerPass.php
+++ b/src/Bootstrap/AddServicesCompilerPass.php
@@ -12,6 +12,8 @@ use Symfony\Component\Yaml\Yaml;
 use Drupal\Console\Utils\TranslatorManager;
 use Drupal\Console\Extension\Extension;
 use Drupal\Console\Extension\Manager;
+use Drupal\Core\Cache\ListCacheBinsPass;
+use GuzzleHttp\Client;
 
 /**
  * FindCommandsCompilerPass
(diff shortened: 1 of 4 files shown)


jmolivas:
@citlacom Thanks for the PR and the great explanation for the changes.

 I tested the code and make some updates I will keep testing my changes and then merge this PR and apply my fixes.

Issue #3599, opened 2017-11-21
[entity:delete] Allow deleting all entities of a type.


cburschka-pwc:
Uninstalling a module via the console frequently fails because the module has created content entities, which need to be deleted:

     [ERROR] The following reasons prevent the modules from being uninstalled:
        There is content for the entity type: @entity_type. <a href=":url">Remove @entity_type_plural</a>.  

It would be great if this deletion could be performed via the console as well. Currently `entity:delete` only deletes a single entity.

Proposed solutions:

1. Add a new command, `entity:delete:all <type>`.
2. Add an optional argument so that `entity:delete <type> --all` deletes all entities of the given type.

cburschka-pwc:
(Merged)

PR #3604, opened 2017-11-23
[entity:delete] Add --all option. (#3599)

cburschka:
Add `--all` to `entity:delete` (#3599).


enzolutions:
Hi @cburschka 

I understand your motivation, but for massive deletion I would recommend to ask a extra confirmation in execution, just to avoid quick fingers and quick regrets.

Kind regards

cc @jmolivas

PR #3632, opened 2017-12-15
offer module as an interactive option in config:export:single - fixes #3631

### LINKED ISSUE #3631: [config:export:single] should offer module as an option
### Problem/Motivation
When using `config:export:single` interactively, it would be helpful to offer `module` as an option.

malcomio:
offer module as an interactive option in config:export:single - fixes #3631


jmolivas:
@malcomio Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3678, opened 2018-01-10
ReMove default assignment alias in abstract class. The alias should b…

LOBsTerr:
…e set in the classes which extend the FormCommand


jmolivas:

PR #3690, opened 2018-01-12
[ajax:command] Fix translation for generate ajax command.

miguel303:
**Changes**
- Use translation from generate.ajax.command rather than generate.controller


jmolivas:
@miguel303 Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3712, opened 2018-01-20
[console] Add new trait ArrayInputTrait, which will parse input nested data.

LOBsTerr:
1) Add new trait ArrayInputTrait, which will parse nested data.
2) Switch multi value options to array to handle them properly, when receive params through the options
3) Fix the methods, which parses the key:value data
4) Fix the logic when we don't use interaction mode.


jmolivas:

PR #3713, opened 2018-01-20
[generate:breakpoint] Add support of multiple values for breakpoints option, when we use inline options.

LOBsTerr:


jmolivas:

PR #3727, opened 2018-01-22
Add symfony/thanks Composer plugin

nicolas-grekas:
I propose to add https://github.com/symfony/thanks to the requirements of Drupal Console so that Drupal users can have the opportunity to easily discover the tool, and can use it to send stars to Drupal also, if they want to.


pfrenssen:
Is this an actual requirement for the project? Isn't it more appropriate to add this as a suggestion than as a requirement?

jmolivas:
I will relocate as `require-dev` to make this package available for development and make easy to skip installing it in production.

nicolas-grekas:
Note that dev deps are not installed when drupal console is installed using composer.
This will make the plugin out of reach for most people.

jmolivas:
@nicolas-grekas true because of `drupal/console` is a dependency of `drupal`. Ideally, this package should be a `require-dev` of Drupal. I will create an issue for that at d.o. and paste the link here. For now, let's keep the dependency as it is.

PR #3745, opened 2018-01-25
[console] Role improvements. Fix #3729 

### LINKED ISSUE #3729: [role:delete] enhancement
### Issue title

1) Role delete command doesn't validate passed role names
2) When we remove the role, it would be nice to have a confirmation.

### Problem/Motivation
We will improve user experience and avoid the errors.

### How to reproduce
Try to enter a fake role name

`drupal role:delete dgdfg`

**Details to include:**
Console verion: 1.5.0

LOBsTerr:
1) Improve validation of the roles
2) Add confirmation
3) Rename ConfirmGeneration to generic ConfirmOperation

### DIFF
--- a/src/Command/Shared/ConfirmationTrait.php
+++ b/src/Command/Shared/ConfirmationTrait.php
@@ -37,4 +37,12 @@ trait ConfirmationTrait
 
         return $confirmation;
     }
+
+    /**
+     * @deprecated
+     */
+    public function confirmGeneration()
+    {
+        return $this->confirmOperation();
+    }
 }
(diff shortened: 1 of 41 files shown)


src/Command/Shared/ConfirmationTrait.php
```diff
@@ -18,7 +18,7 @@
      *
      * @return bool
      */
-    public function confirmGeneration()
+    public function confirmOperation()
```
jmolivas: @LOBsTerr could we add this method to avoid breaking contrib modules using it
```
    /**
     * @deprecated
     */
    public function confirmGeneration()
    {
        return  $$this->confirmOperation();
    }
```

LOBsTerr: Yes, it is true. I will do it

jmolivas: Awesome thanks

LOBsTerr: Done

PR #3747, opened 2018-01-26
A new format output for adding new fields

LOBsTerr:


src/Command/Shared/FormTrait.php
```diff
@@ -40,9 +40,16 @@ public function formQuestion()
...
 
+            $this->getIo()->writeln(sprintf(
+              $this->trans('commands.common.messages.available-field-types'), implode(', ', $input_types)
+            ));
+            $this->getIo()->newLine();
+
             $inputs = [];
             $fieldSets = [];
             while (true) {
+                $this->getIo()->info($this->trans('commands.common.questions.inputs.new-field'));
```
jmolivas: @LOBsTerr Can we use `comment`  instead of  `info`? 

To show the message in yellow color instead to make clear is not another question?

LOBsTerr: @jmolivas Sure, it is a good idea

LOBsTerr:
@jmolivas replace info with comment

jmolivas:

PR #3757, opened 2018-02-01
Small bugfix for `update:execute`, removing extra unneeded parameter …

nvaken:
…from `runUpdates`.

This should fix hechoendrupal/drupal-console#3756


LOBsTerr:
@nvaken Thank you for contribution

Issue #3541, opened 2017-10-04
Use chains to replace site:mode command


marcelovani:
### Issue title

[site:mode] Use chains to replace site:mode command

### Problem/Motivation
Since chains are getting more powerful, I figured we could replace the code found in https://github.com/hechoendrupal/drupal-console/blob/master/src/Command/Site/ModeCommand.php with a chain

**Details to include:**
We can simply have less code to maintain and add more flexibility to users

### Solution
- We should be able to delete most of the code found in
https://github.com/hechoendrupal/drupal-console/blob/master/src/Command/Site/ModeCommand.php and the actual site.mode.yml
- We still need to port the functionality of doing the **services** bit found in site.mode.yml
- We can move the ModeCommand into a folder called Legacy, for people running site:mode command we would tell them to use the new chains
- This issue requires #3540 

This is how the site:mode:dev chain would look
```
command:
  name: site:mode:dev
  description: 'Switch system performance configuration'
commands:
  # system.performance
  - command: config:override
    arguments:
      name: system.performance
        - cache.page.use_internal: false
        - css.preprocess: false
        - css.gzip: false
        - js.preprocess: false
        - js.gzip: false
        - response.gzip: false
  # views.settings
  - command: config:override
    arguments:
      name: views.settings
        - ui.show.sql_query.enabled: true
        - ui.show.performance_statistics: true
  # system.logging
  - command: config:override
    arguments:
      name: system.logging
        - error_level: all
```

The prod chain would be the inverse of the chain above

marcelovani:
Another proof of concept: Use twig + a new command to override services.

```
# How to use
# site:mode --mode="dev"
command:
  name: site:mode
  description: 'Changes the site mode'
vars:
  mode:
    - dev
    - prod
commands:
  # System Performance
  {% set keys = ['cache.page.use_internal', 'css.preprocess', 'css.gzip', 'js.preprocess', 'js.gzip', 'response.gzip'] %}
  {% for key in keys %}
  - command: config:override
    arguments:
      name: system.performance
      key: {{ key }}
      value: {% if mode == "dev" %} false {% else %} true {% endif %}
  {% endfor %}

  # Views settings
  {% set keys = ['ui.show.sql_query.enabled', 'ui.show.performance_statistics'] %}
  {% for key in keys %}
  - command: config:override
    arguments:
      name: views.settings
      key: {{ key }}
      value: {% if mode == "dev" %} false {% else %} true {% endif %}
[...]

PR #3771, opened 2018-02-08
#3541 Removed site:mode command, Added service:override command

marcelovani:


enzolutions:
@marcelovani Muito Obrigado.

We want to include this in the new version, could you include more information about why we need to remove the site:mode and how service:override will replace this command in a more effective way.

Thanks in advance.

Muito Obrigado

marcelovani:
This pr is for this issue https://github.com/hechoendrupal/drupal-console/issues/3541
I did this a long time ago and I can't remember the whole thinking behind it, but I think the idea was to use the service:override to replace existing logic in the site:mode command.

PR #3772, opened 2018-02-08
[config:override] Option to allow nested configs

marcelovani:


src/Command/Config/OverrideCommand.php
```diff
@@ -53,14 +53,19 @@ protected function configure()
...
                 $this->trans('commands.config.override.arguments.key')
             )
             ->addArgument(
                 'value',
-                InputArgument::REQUIRED,
+                InputArgument::OPTIONAL,
                 $this->trans('commands.config.override.arguments.value')
             )
+            ->addArgument(
+                'key-values',
```
jmolivas: @marcelovani can we use as singular `'key-value`

marcelovani:
Done

enzolutions:
@marcelovani could you provide an example config with nested values to test, please

marcelovani:
To be honest I don't remember the use case.

Maybe something like https://git.drupalcode.org/project/drupal/raw/8.8.x/core/modules/image/config/install/image.style.large.yml

```
langcode: en
status: true
dependencies: {  }
name: large
label: 'Large (480×480)'
effects:
  ddd73aa7-4bd6-4c85-b600-bdf2b1628d1d:
    uuid: ddd73aa7-4bd6-4c85-b600-bdf2b1628d1d
    id: image_scale
    weight: 0
    data:
      width: 480
      height: 480
      upscale: false
```

How can I override `effects.data.width`

Maybe it is possible, I don't know, its been a while.

enzolutions:
Thanks @marcelovani for your contribution, was used as base to be solved with PR https://github.com/hechoendrupal/drupal-console/pull/4047

PR #3774, opened 2018-02-09
[generate:plugin:migrate:dataparser] Add new command for generating data parser plugin for migration. Fix #3773

### LINKED ISSUE #3773: [generate:plugin:migrate:data_parser] [New Command] Command to create data parser plugin for migration.
Command to create the data parser plugin for the migration.

joshirohit100:
Issue - #3773

### DIFF
--- a/src/Command/Generate/PluginMigrateDataParserCommand.php
+++ b/src/Command/Generate/PluginMigrateDataParserCommand.php
@@ -15,7 +15,6 @@ use Drupal\Console\Core\Command\ContainerAwareCommand;
 use Drupal\Console\Generator\PluginMigrateDataParserGenerator;
 use Drupal\Console\Command\Shared\ModuleTrait;
 use Drupal\Console\Command\Shared\ConfirmationTrait;
-use Drupal\Console\Extension\Manager;
 use Drupal\Console\Core\Utils\StringConverter;
 use Drupal\Console\Core\Utils\ChainQueue;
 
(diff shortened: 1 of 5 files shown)


src/Command/Generate/PluginMigrateDataParserCommand.php
```diff
@@ -0,0 +1,174 @@
...
+        $this->extensionManager = $extensionManager;
+        $this->stringConverter = $stringConverter;
+        $this->validator = $validator;
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $this
+            ->setName('generate:plugin:migrate:data_parser')
```
LOBsTerr: @joshirohit100 Normally, underscore is not used in command name.
some examples generate:plugin:rulesaction, generate:plugin:rulesaction and etc.

Maybe it would be better to use something like this generate:plugin:migrate:data:parse. So, if we add also data fetcher we will have generate:plugin:migrate:data:fetcher. What do you think ?

joshirohit100: I am fine with both generate:plugin:migrate:dataparser or generate:plugin:migrate:data:parse
Lets see what other say

jmolivas: I like `generate:plugin:migrate:dataparser` better

joshirohit100: Thanks! I have updated the PR

joshirohit100:
@jmolivas waiting for this :)

Issue #3788, opened 2018-02-15
Move symfony/thanks from a required dependency to a suggestion


pfrenssen:
### Problem/Motivation

In #3727 the symfony/thanks project was added as a dependency to Drupal Console. This project is a nice initiative to elicit some response and additional exposure from the PHP community.

However adding this as a requirement to Drupal Console will cause this software to be installed unsolicited on tens of thousands of development machines and servers.

For small projects and personal hobby sites this might not be a problem, but Drupal is also used for large corporate and government projects that have strict policies regarding the software that can be used. It is very hard to convince management why we suddenly require to have software on our servers that is doing nothing but distributing "internet points" in a fun way.

### Solution

I think the best way would be to use the `suggest` feature of Composer to add symfony/thanks to the project. This will still expose the package to everybody who uses drupal-console, but does not make it a hard requirement. Developers can then decide for themselves whether to include it or not.

dunglas:
If you really have such strict restrictions, you can just add this snippet in your composer.json:

```
    "replace": {
        "symfony/thanks": "*"
    },
```

nicolas-grekas:
Related to https://github.com/drupal-composer/drupal-project/issues/352
I don't adhere to the reasoning presented here.
Adding Symfony/thanks or drupal-console to the requirements create an experience that is an integral part of what the authors want to provide. Code is never cold and purely technical.

JorisVanEijden:
I don't see how github stars can be an integral part of a development/maintenance console experience.

You are also adding a *dependency*. That means that if ever something goes wrong with "thanks" then drupal console will break.

Use `require` for the bare minimum absolute necessities that your package simply won't work without and use `suggest` for providing an optimal "experience".

jmolivas:
@alexpot Adding a link to the d.o. issue `Add symfony/thanks to Drupal` by `alexpott` to keep track of replies => https://www.drupal.org/project/drupal/issues/2945263

pfrenssen:
> If you really have such strict restrictions, you can just add this snippet in your composer.json:
>
>     "replace": {
>         "symfony/thanks": "*"
>     },

I really don't think that forcing people to mess with the requirements of their dependencies can ever be considered a good practice.

Chi-teck:
[...]

PR #3789, opened 2018-02-15
Make symfony/thanks a suggestion rather than a hard requirement

pfrenssen:
Fix for #3788

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,8 +50,10 @@
         "symfony/css-selector": "~2.8|~3.0",
         "symfony/dom-crawler": "~2.8|~3.0",
         "symfony/expression-language": "~2.8|~3.0",
-        "symfony/http-foundation": "~2.8|~3.0",
-        "symfony/thanks": "^1.0"
+        "symfony/http-foundation": "~2.8|~3.0"
+    },
+    "suggest": {
+        "symfony/thanks": "Thank your favorite PHP projects on Github using the CLI!"
     },
     "bin": ["bin/drupal"],
     "config": {


nicolas-grekas:
This doesn't provide the same experience at all.
symfony/thanks is a requirement to the experience the authors of drupal-console want to provide out of the box. Thus it belongs to "require". I understand that you have a different definition of what "require" means, but there is nothing anywhere that makes your definition more accurate than this definition given here: a requirement to the experience. That's a great thing.

dkarlovi:
Perhaps a better project to provide that experience would be Composer? This avoids needing to do it for each and every project separately.

nicolas-grekas:
@dkarlovi see https://github.com/symfony/thanks/issues/35

dkarlovi:
@nicolas-grekas yeah, I've seen that issue / tweet before. My point is, if Composer doesn't want to have it be part of their default experience (where it IMO makes way more sense), why should random projects?

Not trying to imply the intention is bad in any way, on the contrary. Just think that adding a Composer plugin as a **base requirement** to random projects is not a good way forward from a technical standpoint.

nicolas-grekas:
> Composer doesn't want to have it be part of their default experience

that's certainly not the reason why this is not in Composer. The only reason that is in this tweet in the maintenance burden this would create on Composer. That's a perfectly valid point. I'm way more interested in maintaining this plugin myself, so that Composer guys can focus their time on core Composer features.

dkarlovi:
> maintenance burden this would create on Composer

That's "Composer doesn't want it" said in a different way. :smile: 

How do you reconcile advocating to add this here as `require` while [Symfony doesn't `require-dev` code it actually references](https://github.com/symfony/symfony/pull/25536#discussion_r157591181)? Seems like a double standard.

Anyway, I understand you're on a mission to get this across (and IMO it a valuable cause), but still think there's got to be a better way to do this.

davereid:
There's no way to disable the message. I'm seeing it every time that composer update pulls in new packages, and for a large project that is often. I get it, but now I have no way to NOT see it.

jmolivas:
@pfrenssen Thanks for your contribution, your PR was merged. This will be included in the next release.

pfrenssen:
Thanks!

PR #3800, opened 2018-02-21
[console] remove drupal/console-dotenv dependency.

jmolivas:


hussainweb:
I noticed this just now on performing an update. I'd appreciate it if you can share some background on why this was done.

jmolivas:
@hussainweb  If using `.env` files, the `vlucas/phpdotenv` dependency must be installed manually using composer.

composer require vlucas/phpdotenv --dev

jmolivas:
Since `vlucas/phpdotenv` is a dev dependency. Should be added via the main composer.json file.
If using docker and properly configuring the services on your `docker-compose.yml` file then those `.env` variables are registered as system ENV variables.

PR #3822, opened 2018-03-13
3275 remove translation code

LOBsTerr:
Remove translations related code if we didn't choose entity as translatable


Alluuu:
Is langcode required, or I can manually remove it if I wish to? (I'm slightly confused by the comment of it being default)

LOBsTerr:
@Alluuu I removed the langcode at the previous commit, then I realized that this is base field, there are no reasons to remove it and put it back
More information is here: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21ContentEntityBase.php/function/ContentEntityBase%3A%3AbaseFieldDefinitions/8.2.x

PR #3846, opened 2018-03-24
config:export:content:type -> added 'all' option for exporting all content types

novia713:
captura de pantalla 2018-03-24 a las 17 06 08

if passing content type directly as argument it's possible now:
`../vendor/bin/drupal config:export:content:type all --module="my_module"`


enzolutions:
@novia713 thanks for your contribution, the conflict and some missing pieces were solved on https://github.com/hechoendrupal/drupal-console/pull/4036 by @harold20

PR #3867, opened 2018-04-17
[template] Replace deprecated drupal_set_message

bahuma20:
drupal_set_message is marked as deprecated and will be removed in drupal 9.

We should use best practices in the templates and fix deprecation issues as fast as possible to keep the effort needed to port this project to drupal 9 minimal.


jmolivas:
@bahuma20 Thanks for your contribution, your PR was merged. This will be included in the next release.

PR #3871, opened 2018-04-22
[multisite:new] fix so that multi site command uses uri argument

t14:
## Problem/Motivation
When running the `drupal multisite:new ` the second argument `uri` is never used and the `directory` argument is used in its place making it impossible to take full advantage of all the functionality in the `sites.php` file that this command writes to

## How to reproduce
run `Drupal multisite:new new-site http://new-site.example.com`
In `sites.php` you will see `$sites['new-site'] = 'new-site';`
When the final result should actually be `$sites['http://new-site.example.com'] = 'new-site';`

## Solution
The solution in this PR fixes the issues

### DIFF
--- a/src/Command/Multisite/NewCommand.php
+++ b/src/Command/Multisite/NewCommand.php
@@ -154,7 +154,7 @@ class NewCommand extends Command
             throw new FileNotFoundException($this->trans('commands.multisite.new.errors.sites-missing'));
         }
 
-        $sites_file_contents .= "\n\$sites['$this->directory'] = '$this->directory';";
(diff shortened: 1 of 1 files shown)


harold20:
Hi @t14 thanks for this PR.

Only one question. Is it practical to have a different name between directory and uri?  because with your change this can happen.

harold20:
I just create this PR @enzolutions  https://github.com/hechoendrupal/drupal-console/pull/4039 to validate the url name, because I had a problem with the console if I define the uri name starting with `http://`

enzolutions:
Thanks @t14

Issue #3845, opened 2018-03-23
[console] alchemy/zippy version


Pasqualle:
Why is the alchemy/zippy library hard-coded on the 0.4.3 release?

As I see it was last updated in https://github.com/hechoendrupal/drupal-console/pull/3053 , but I did not find any comment about why it is set to exact version number.
Can we just change the requirements to "^0.4"?

This is the only package in my Drupal install which is outdated (at patch release).

mike-potter:
I have also run into a problem with this.  This hard-coded version causes drupal/console to conflict with any of our projects using Pattern Lab.  Pattern Lab uses "^0.3".  This is it's own problem that I have reported because for versions less than 1.0 using "^0.3" prevents composer from using v0.4.  So even setting drupal/console to ^0.4 would continue to conflict.

Is there a reason the exact 0.4.3 or 0.4 version is needed in Console?  If there is no specific dependency on new functionality, this should be changed to "~0.3" which would support both existing 0.3 versions from Pattern Lab, but would also support the latest 0.4 version.  (Using ^ has special behavior for versions less than 1.0 whereas ~ will allow any 0.x version).

LOBsTerr:
I have tested it. It woks perfectly with both 0.3 ++ and 0.4 ++ versions. So, we can user ~0.3 as it was suggested by Mike.

PR #3876, opened 2018-04-24
allow alchemy/zippy version to go above 0.4.3

artiswilliams:
This addresses the following issue: #3845 

Unless there was a reason to lock to a specific version, then ignore.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
     },
     "require": {
         "php": "^5.5.9 || ^7.0",
-        "alchemy/zippy": "0.4.3",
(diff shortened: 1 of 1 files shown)


LOBsTerr:
@artiswilliams Can you change to ~0.3. As it was suggested in the related issue. I have tested, it works fine with both 0.3++ and 0.4++ versions with DC.
Thanks

dpagini:
This has been open for a long time. Just came across this same issue and would love to see it go in. Since the OP hasn't changed anything for ~1 year, would I be OK to just open a new PR with the requested changes?

PR #3880, opened 2018-05-01
Added Paragraph type export command

joe-mcpherson:
Added Paragraph type export to export Paragraph type's fields with following command:

drupal config:export:paragraph:type

(Translation in separate pull request on drupal-console-en repository)

### DIFF
--- a/config/services/config.yml
+++ b/config/services/config.yml
@@ -24,6 +24,11 @@ services:
     arguments: ['@entity_type.manager', '@config.storage', '@console.extension_manager', '@console.validator']
     tags:
       - { name: drupal.command }
+  console.config_export_paragraph_type:
+    class: Drupal\Console\Command\Config\ExportParagraphTypeCommand
+    arguments: ['@entity_type.manager', '@config.storage', '@console.extension_manager', '@console.validator']
+    tags:
+      - { name: drupal.command }
   console.config_export_single:
     class: Drupal\Console\Command\Config\ExportSingleCommand
     arguments: ['@entity_type.manager', '@config.storage', '@console.extension_manager','@language_manager', '@console.validator']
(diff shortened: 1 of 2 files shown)


LOBsTerr:
@jmolivas I propose universal solution https://github.com/hechoendrupal/drupal-console/pull/3875. This will reduce amount of code we have to support. (https://github.com/hechoendrupal/drupal-console/issues/3794)

If we still need export of entity as separate command we should use chain instead, but still I don't think it will be useful, if we have a generic command

joe-mcpherson:
@LOBsTerr I agree a generic solution makes more sense. I didn't see your proposal, I just needed this for a project I was working on so thought I'd throw it back if it could be useful for someone

jmolivas:
@LOBsTerr I have to say I like those commands to export configurations and use the result on a custom profile.

I also agree a generic solution one single command could be easy to maintain since is one command but maybe the code gets a little complex since is taking a lot of considerations.

LOBsTerr:
@jmolivas I'm absolutely agree with and in fact I have just a little bit extended the current code and I moved the common classes to the trait ExportTrait.php. Check the PR https://github.com/hechoendrupal/drupal-console/pull/3875/files

anuragshah17:
@joe-mcpherson Hey i am trying to apply patch for this pull, not working with Drupal Console 1.0.1 version. I need to try this with dev version for Drupal Console?

anuragshah17:
@joe-mcpherson Please ignore above message. Command worked with 1.8.0 version. 
Thanks

LOBsTerr:
I was thinking why we limit us only to specific config entities and I added a new commit, which expose all config entities, because I have seen recently request about media type also. Please, test it https://github.com/hechoendrupal/drupal-console/pull/3875

LOBsTerr:
I have merged https://github.com/hechoendrupal/drupal-console/pull/3875
There you can export any config entity

Issue #3508, opened 2017-08-31
[multisite:new]  argument 'uri' is un-used.


mattsqd:
### Problem/Motivation
[multisite:new documentation](https://docs.drupalconsole.com/en/commands/multisite-new.html)
When passing the second argument to the second parameter, it is unused by src/Command/Multisite/NewCommand.php. What happens is that the first argument, the sites directory, is used for both key and value of sites.php.

**Details to include:**
- I believe a change was made [here](https://github.com/hechoendrupal/drupal-console/issues/3197#issuecomment-287102291), that I believe was because the confusing term 'uri' is used in this command that is different than other commands.
- URI sometimes means the URL to reach the site https://test.com, the URI as in test.com, but in this case, shouldn't it be keys like seen in this [example](https://api.drupal.org/api/drupal/sites%21example.sites.php/8.2.x)?

### How to reproduce
drupal multisite:new my_site_dir my_site_uri
When finished, sites.php should contain an entry with my_site_dir as key and value.

**Details to include:**
- Drupal version 8.3.7
- Console version 1.0.1.
- Console Launcher version 1.0.1.

### Solution
Revert [this commit ](https://github.com/hechoendrupal/drupal-console/commit/c38ae931144b2144bf72b071c97f420486c1d9d2)

Update the documentation to reference how keys of sites.php are created.

cburschka-pwc:
I'm struggling to understand the purpose of that commit - mapping the directory name to itself is obviously not correct, and I don't see how it fixes anything it was supposed to fix.

However, the original code was also wrong, and it doesn't seem trivial. Just as you say: The correct syntax of the array key in `$sites` is not a URI, but a dotted string mask, with various possible values.

I suspect the best fix would be reverting the bad commit, and renaming the "URI" to a more accurate "mask" argument: `drupal multisite:new <directory> <mask>`

As a convenient feature, if no mask is given, the command could interactively list suggestions based on the value of a `--uri` option.

Eg: `drupal multisite:new mysite www.mysite.com` does exactly that, while `drupal multisite:new mysite --uri https://www.mysite.com:8888` does something like the following:

```
$ drupal multisite:new mysite --uri https://www.mysite.com:8888/test/

Choose one of the following masks:

[0] 8888.www.mysite.com.test
[1] 8888.www.mysite.com
[2] www.mysite.com.test
[3] www.mysite.com
[4] mysite.com
```

[...]

Issue #3197, opened 2017-02-21
[chain] - [multisite:new] [site:install] - Executing chain fails


brandonratz:
**Version:** Drupal Console (1.0.0-rc16)

# Problem
Using command: 'multisite:new' followed by 'site:install' via chain command fails.

# Expected behavior
1. Creates multisite directory
2. Installs site via multisite uri

# Actual Behavior

1. Creates multisite directory
2. ERROR: Invalid multisite error

## Console Output

```
 // multisite:new

[OK] The new multisite structure was successfully created at "sites/test.example.com" and is ready for installation.

 // site:install

[ERROR] Invalid multisite, please create multisite using command drupal multisite:new test.example.com
         --site-uri=test.example.com`
```

## Chain file example

```
commands:
  - command: 'multisite:new'
    options:
        learning: true
        uri: 'http://test.example.com'
    arguments:
        directory: 'test.example.com'
        uri: 'http://test.example.com'
  - command: 'site:install'
    options:
        langcode: en
        db-type: mysql
        db-host: localhost
        db-name: test_example_dev
        db-user: root
        db-pass: root
        db-port: '3306'
        site-name: 'Example Website'
        site-mail: admin@example.org
        account-name: admin
        account-mail: admin@example.org
        account-pass: password
        learning: true
        uri: 'http://test.example.com'
    arguments:
        profile: lightning
```

michaellenahan:
Hello there! I met @enzolutions in the sprint room at Drupalcamp Iceland, and asked if there was somehow I could help to get started contributing to DrupalConsole.

He pointed me to this ticket. I'm saying this to warn you I'm a newbie to contributing here ...

I have cloned drupal-console to ~/workspace/drupal-console and the drupal site I am running it from is ~/workspace/drupal.template.

When I run the command with the supplied chain file, I get the following output.
```
$ cd ~/workspace/drupal.template/.
$ ~/workspace/drupal-console/bin/drupal chain --file=~/workspace/chain.yml --debug
✔ Creating request
➤ Creating Drupal kernel


  [Symfony\Component\Console\Exception\CommandNotFoundException]
  The command "multisite:new" does not exist.
```

michaellenahan:
I'm not sure why I'm getting:
```
The command "multisite:new" does not exist.
```
After all I can see the multisite:new command in the Repo. (It was added with this PR https://github.com/hechoendrupal/drupal-console/pull/2335/files)

brandonratz:
[...]

PR #3889, opened 2018-05-17
Multisite new with uri

rpsu:
Fix what got broken with #3197 and which causes problems like #3508


enzolutions:
Fixed with https://github.com/hechoendrupal/drupal-console/pull/4039 and https://github.com/hechoendrupal/drupal-console/pull/3871

Issue #3898, opened 2018-06-04
[user:password:reset] Notice when reseting user password


FireGhost:
### Problem/Motivation
When I want to reset the admin password using `drupal upr admin`, a notice appears:

```
 [ERROR] <em class="placeholder">Notice</em>: Only variables should be passed by reference in <em
         class="placeholder">Drupal\Console\Command\User\UserBase-&gt;getUserEntity()</em> (line <em
         class="placeholder">48</em> of <em
         class="placeholder">vendor/drupal/console/src/Command/User/UserBase.php</em>). <pre
         class="backtrace">Drupal\Console\Command\User\UserBase-&gt;getUserEntity(&#039;admin&#039;) (Line: 91)
         Drupal\Console\Command\User\PasswordResetCommand-&gt;execute(Object, Object) (Line: 242)
         Symfony\Component\Console\Command\Command-&gt;run(Object, Object) (Line: 861)
         Symfony\Component\Console\Application-&gt;doRunCommand(Object, Object, Object) (Line: 193)
         Symfony\Component\Console\Application-&gt;doRun(Object, Object) (Line: 185)
         Drupal\Console\Core\Application-&gt;doRun(Object, Object) (Line: 64)
         Drupal\Console\Application-&gt;doRun(Object, Object) (Line: 117)
         Symfony\Component\Console\Application-&gt;run() (Line: 89)
         require(&#039;/var/www/html/vendor/drupal/console/bin/drupal.php&#039;) (Line: 4)
         </pre>
```

### How to reproduce
Drupal console: **1.8.0**
PHP: 7.0.29
Drupal: 8.3.9

Simply do a `drupal user:password:reset admin` in the command line.
The error doesn't appear if I use the UID (admin => 1): `drupal upr 1`.

### Solution
The problem comes from the line 49 of the file UserBase.php. On this line, it's trying to use `reset()` on a method. But the function `reset()` needs a variable as a parameter, not a method.

I'm creating a pull request right now.

LOBsTerr:
@FireGhost Thank you for your contribution. It was merged

PR #3899, opened 2018-06-04
Fix issue #3898 [user:password:reset] Notice when reseting the user password

FireGhost:
This PR fix the issue #3898


marcelovani:
Do you know when we can have a release that includes this fix?

PR #3903, opened 2018-06-28
Test PHP 7.2 and 7.3 and support 7.x

hugovk:
* PHP 7.2 was released in November 2017, so test it too.
* PHP 7.3 was released in December 2018, so test it too.
* Support PHP 7.x by not allowing them to fail.


hugovk:
Rebased on master to keep current.

hugovk:
Updated to add PHP 7.3.

enzolutions:
Thanks @hugovk

PR #3232, opened 2017-03-16
Fix contents of generated sites.php file (#3197)

michaellenahan:

PR #3926, opened 2018-08-24
fix: use user supplied uri back

CaptainQuirk:
## :question: What does this PR do ?

It kinda reverts #3232 (#3197) and follows on #3508 .

It is my conviction that a user of the `multisite:new` command should know the different ways an site alias can be written.

Those different ways are listed in the `sites/example.sites.php` of a regular **drupal 8** installation.


enzolutions:
Thanks @CaptainQuirk for your contribution, already implemented.

PR #3947, opened 2018-10-12
Keep current maintenance mode status.

GuidoRobertone:


LOBsTerr:
@GuidoRobertone Thanks for your contribution, but it is not a proper way to do that.
We have a configuration, which will enable the maintenance mode before the execution of a command and restore it back after
Check this:
https://github.com/hechoendrupal/drupal-console/blob/912508f5465962877659480198802c43231e1057/src/Command/Database/RestoreCommand.php#L60

Can you please fix it

LOBsTerr:
@GuidoRobertone Thank you for your contribution. Closing this one. I have created  a new one https://github.com/hechoendrupal/drupal-console/pull/3963

PR #3960, opened 2018-10-19
1088 interfaces into src

LOBsTerr:


bojanz:
Merging this with 0 community feedback was unwise and makes Console's entity templates one step closer to unusable. I'm afraid I won't be able to recommend it anymore.

PR #3996, opened 2019-01-13
[update:execute] run trailing post updates if there is no updates

lalop:
We can have trailing postUpdates to run or entities update without having any module update.
This pr try to fix that

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "composer/installers": "~1.0",
         "doctrine/annotations": "^1.2",
         "doctrine/collections": "^1.3",
-        "drupal/console-core": "dev-master",
+        "drupal/console-core": "1.8.0",
         "drupal/console-extend-plugin": "~0",
         "guzzlehttp/guzzle": "~6.1",
         "psy/psysh": "0.6.* || ~0.8",
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "composer/installers": "~1.0",
         "doctrine/annotations": "^1.2",
         "doctrine/collections": "^1.3",
-        "drupal/console-core": "1.8.0",
+        "drupal/console-core": "dev-master",
         "drupal/console-extend-plugin": "~0",
         "guzzlehttp/guzzle": "~6.1",
         "psy/psysh": "0.6.* || ~0.8",
--- a/src/Command/Update/ExecuteCommand.php
+++ b/src/Command/Update/ExecuteCommand.php
@@ -108,7 +108,8 @@ class ExecuteCommand extends Command
                 $this->trans('commands.update.execute.options.update-n'),
                 '9000'
             )
-            ->setAliases(['upex']);
+            ->setAliases(['upex'])
+            ->enableMaintenance();
     }
 
     /**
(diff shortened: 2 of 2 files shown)


lalop:
do you think this can be merged ?

composer.json
```diff
@@ -41,7 +41,7 @@
         "composer/installers": "~1.0",
         "doctrine/annotations": "^1.2",
         "doctrine/collections": "^1.3",
-        "drupal/console-core": "1.8.0",
```
LOBsTerr: Can you revert this changes ? The version should be fixed

src/Command/Update/ExecuteCommand.php
```diff
@@ -151,42 +154,49 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                return 1;
+            }
         }
+        
 
+        // Post Updates are only safe to run after all schemas have been updated.
         if (!$this->getUpdates()) {
+            $postUpdates = $this->postUpdateRegistry->getPendingUpdateInformation();
+            
+            if(!empty($postUpdates)) {
```
LOBsTerr: Space is missing

src/Command/Update/ExecuteCommand.php
```diff
@@ -151,42 +154,49 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                watchdog_exception('update', $e);
+                $this->getIo()->error($e->getMessage());
+                return 1;
+            }
         }
+        
 
+        // Post Updates are only safe to run after all schemas have been updated.
         if (!$this->getUpdates()) {
+            $postUpdates = $this->postUpdateRegistry->getPendingUpdateInformation();
```
LOBsTerr: It looks like you set maintenance mode, but it will be reverted only when where are $postUpdates. In other case the website will stay in maintenance mode.
Actually, there is better mechanism for  maintenance mode in the Console core, it is merged, but it is not tagged yet. I asked to create a new tag.

src/Command/Update/ExecuteCommand.php
```diff
@@ -271,9 +281,8 @@ private function runUpdates(
     /**
      * @return bool
      */
-    private function runPostUpdates()
+    private function runPostUpdates($postUpdates)
```
LOBsTerr: Based on the comment above, we can keep it as it was

src/Command/Update/ExecuteCommand.php
LOBsTerr: Check the way we will activate maintenance mode

```
protected function configure()
    {
        $this
            ->setName('update:entities')
            ->setDescription($this->trans('commands.update.entities.description'))
            ->enableMaintenance()
            ->setAliases(['upe']);
    }

```

composer.json
lalop: done

src/Command/Update/ExecuteCommand.php
lalop: fixed

src/Command/Update/ExecuteCommand.php
lalop: it looks much better with this new helper.
I had updated the code, do you want me to rebase the pr ?

LOBsTerr:
@lalop Thank you for your work, no need to rebase. I will merge as soon as we set a new tag on the Drupal console core

lalop:
@LOBsTerr do you have any idea of when does drupal console core will be tagged ?

LOBsTerr:
I have merged it will be tagged soon.

PR #4017, opened 2019-04-16
Update composer.json since drupal-console is not compatible with drupal-console-core 1.8.0

ShaunDychko:
`composer.json` lists `drupal/console-core` as a dependency pinned to version 1.8.0. However, this version is missing `\Drupal\Console\Core\Command\Command::enableMaintenance`. This method is present in the `dev-master` version of `drupal/console-core`, so the composer.json in `drupal/console` is updated accordingly in order for `drupal` to run.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "composer/installers": "~1.0",
         "doctrine/annotations": "^1.2",
         "doctrine/collections": "^1.3",
-        "drupal/console-core": "1.8.0",
+        "drupal/console-core": "dev-master",
         "drupal/console-extend-plugin": "~0",
         "psy/psysh": "0.6.* || ~0.8",
         "symfony/css-selector": "~2.8|~3.0",


LOBsTerr:
There is a plan to present a new release soon

mondrake:
Can you please merge this in the meantime?

If you do 
`$ composer require drupal:console:dev-master`
then you end up in an unusable setup (because Composer loads console-core:1.8.0 that does not work with console:dev-master. 

You can't solve it composer-way because you cannot require console-core:dev-master,  so the only way ATM is to override manually console-core with a git clone, breaking the composer workflow :(

With this in you could do `$ composer require drupal:console:dev-master` and get all the latest code, good for any automated testing that uses composer for the codebase build.

enzolutions:
This will be dangerous for production environments, let's wait some days for the new release to see if this fix your issues.

rlmumford:
@mondrake  an alternive workaround that doesn't break the composer workflow would be to us an alias https://getcomposer.org/doc/articles/aliases.md

$ composer require drupal/console-core:"dev-master as 1.8.0" 

This will download the latest dev version by treat it as 1.8.0 for dependency management.

mondrake:
@rlmumford thanks for the tip

Issue #4021, opened 2019-05-07
[config:export:view] option include-module-dependencies not working correctly


stefan-korn:
### Problem/Motivation
Running [config:export:view] with option [include-module-dependencies] set to yes does not work as expected.

What would be expected:
Module dependencies resulting from view will get added to info.yml file

What is actually happening:
info.yml gets overwritten and contains only dependencies resulting from view, everything else is lost from info.yml. This breaks the module effectively.

### How to reproduce
run drupal config:export:view and answer 'Include view module dependencies in module info YAML file" with yes

- Drupal version 8.6.15
- Console version 1.8.0

harold20:
Hi @stefan-korn I made this PR https://github.com/hechoendrupal/drupal-console/pull/4026 that solve this issue

enzolutions:
This change will be included in version 1.9.0

PR #4023, opened 2019-05-08
[config export view] option include-module-dependencies not working

stefan-korn:
hopefully fixing #4021


stefan-korn:
solved with #4026

PR #4045, opened 2019-05-16
4044 config export all option

LOBsTerr:
1) Add export all option like it was done in content type export command
2) Replace content type export command with chainQueue which will call Entity export command inside
3) Replace view export command with chainQueue which will call Entity export command inside
4) Add include dependencies option for content type export command and for entity export commands
5) Small code refactoring
6) Fix issue with HASH and UUID removal for some configuration
7) Replace option all with "-all-", because we can have bundle name all


harold20:
Hi @LOBsTerr thanks for the PR.

I was reviewing the PR and I find an issue. If I choose the option ```all```, all the config to export within  ```$this->configExport``` are generated in each content type 

demo

To fix it, you need to empty ```$this->configExport``` after exporting each content type

image

I hope you can fix it to merged. :)

LOBsTerr:
@harold20 Thanks for review. Nice catch. It has been fixed

enzolutions:
Thanks @LOBsTerr and @harold20

PR #4046, opened 2019-05-19
[site:status] Use Database::getConnectionInfoAsUrl to get the db URL

mondrake:
Use Drupal's method to determine the db connection's URL, instead of own logic. Hide the password if existing, for security reasons.


enzolutions:
@mondrake thanks for your contribution

PR #4049, opened 2019-05-21
[site:install] Improve support of --db-url

mondrake:
Looking at the newly introduced --db-url argument for site:install, I believe there's room for some improvements.

If --db-url is specified **and valid**, there is no purpose to go through the other options. We can just install based on that.

Here's my suggested changes  - **not for merging yet** but rather for sharing with the community - I am new to Drupal console contribution so please bear with me.

My goal: allow `site:install` to run on ANY database driver (also contributed), as long as it complies to the Drupal API for `Database::convertDbUrlToConnectionInfo` and `Database::getConnectionInfoAsUrl`.


enzolutions:
Thank you @mondrake  for your contribution

PR #4051, opened 2019-05-21
[site:install] Enable db url with sqlite

harold20:
We can install a site using sqlite using a database url 
```
drupal site:install standard sqlite://localhost/sites/default/files/.ht.sqlite --no-interaction
```


mondrake:
See also #4049, for an alternative (more generic?) approach

Issue #3929, opened 2018-09-10
[update:execute] Batch updates are not supported


longwave:
### Problem/Motivation

In https://www.drupal.org/project/drupal/issues/2997960 it was identified that `taxonomy_update_8502()` is not run correctly via `update:execute` when more than 100 terms are present in the database.

`hook_update_N` allows long running update operations to be batched via the $context parameter, as documented at https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_update_N/8.6.x but it appears Drupal Console does not implement this, as per https://github.com/hechoendrupal/drupal-console/blob/master/src/Command/Update/ExecuteCommand.php#L338-L340 which only ever runs the update hook once.

### How to reproduce

* Start on Drupal 8.5.x and add over 100 taxonomy terms
* Upgrade to Drupal 8.6.x and run `update:execute`
* Note that only the first 100 taxonomy terms are correctly migrated correctly to the new parent field structure during `taxonomy_update_8502()`

### Solution

Implement batch updates during `update:execute`.

### See also

#3574 
#3787

enzolutions:
@rlmumford we could close this issue after merge https://github.com/hechoendrupal/drupal-console/pull/4057 ?

rlmumford:
@enzolutions I believe we can

enzolutions:
Awesome.

PR #4057, opened 2019-05-22
[update:execute] Support batch updates and post update functions

rlmumford:
See Issue #3929

### Problem/Motivation
drupal updb does not support batch updates. This has made updating from 8.6.x - 8.7.0 impossible as several of the post_update functions require batch processing.

Details of how to reproduce can be found on issue #3929 

### Solution
Update ExecuteCommand::executeUpdate() to loop over the update function until it is completed. Provide helpful feedback to the user throughout this process.

I've added a `do { } while() ` loop into executeUpdate to ensure that the function get's run the correct number of time.

A default message of 'Processed x%' is printed as information and indented to make clear that it relates to the function above.

batchupdatefunction

### DIFF
--- a/src/Command/Update/ExecuteCommand.php
+++ b/src/Command/Update/ExecuteCommand.php
@@ -348,13 +348,24 @@ class ExecuteCommand extends Command
 
     private function executeUpdate($function, &$context)
     {
-        if (!$context || !array_key_exists('sandbox', $context)) {
-            $context['sandbox'] = [];
-        }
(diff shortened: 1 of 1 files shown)


enzolutions:
Thanks, @rlmumford for your contribution.

Do you think you can help us with these issues related https://github.com/hechoendrupal/drupal-console/issues/3929 , https://github.com/hechoendrupal/drupal-console/issues/3574 and https://github.com/hechoendrupal/drupal-console/issues/3787

PR #4061, opened 2019-05-23
[site:install] Small improvements

mondrake:
1. If Drupal is already installed, and --force is not set, there's no point to continue an interactive install that will lead to an error anyway. Exit early.
2. The Drupal installer, when erroring out, returns HTML, not good for displaying on CLI. Cleanup from tags and entities before outputting.
3. A trivial bug fix at lines 271-272
4. Few doc cleanups

### DIFF
--- a/src/Command/Site/InstallCommand.php
+++ b/src/Command/Site/InstallCommand.php
@@ -180,6 +180,14 @@ class InstallCommand extends ContainerAwareCommand
      */
     protected function interact(InputInterface $input, OutputInterface $output)
     {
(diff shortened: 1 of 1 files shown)


enzolutions:
Hi @mondrake 

Thanks for your contribution. I just followed the issue that you opened https://www.drupal.org/project/drupal/issues/3056698, I will check again when that issue get resolved in Drupal Core

mondrake:
Thanks @enzolutions 

Anyone watching, reviews @ https://www.drupal.org/project/drupal/issues/3056698 appreciated.

PR #4077, opened 2019-06-04
Introduce an html to text facility

mondrake:
Drupal outputs HTML text in some cases, which is not fit to present in a CLI environment (example, in site:status and site:install). In Drupal console there are places where `strip_tags` and `html_entity_decode` are used to clean up a bit, but any formatting and links are lost.

I suggest to add a dependency to [html2text/html2text](https://packagist.org/packages/html2text/html2text) and build a facility service to help converting HTML strings to a prettier text.

Example 1: site:install error message

before

2019_06_03_19_06_40_raspberrypi_Dataplicity
 
after

2019_06_04_09_03_21_raspberrypi_Dataplicity

Example 2: site:status (verbose)

before

2019_06_04_10_48_47_raspberrypi_Dataplicity

after

2019_06_04_10_49_51_raspberrypi_Dataplicity


harold20:
Hi @mondrake thanks for the PR.
I am trying to review this PR but I am having trouble with `html2text/html2text`

When I execute `drupal site:install` I have the following error

```
Error: Class 'Html2Text\Html2Text' not found in /Users/path-to-console/drupal-console/src/Utils/HtmlConverter.php 
on line 27 #0 /Users/path-to-console/drupal-console/src/Command/Site/InstallCommand.php(587): Drupal\Console\Utils\HtmlConverter::html2text('Resolve all iss...', Array)
```
I have the library installed.
```
drupal-console git:(4077-introduce-html) ✗ composer show | grep html2text/html2text
html2text/html2text                   4.2.1    Converts HTML to formatted plain text
```

Do you know what it can be?

mondrake:
The class is missing from the autoloader map, maybe you git cloned it instead of indtalling via composer?

Try 

composer require html2text/html2text

harold20:
I can't solve the issue, I installed the dependency on my drupal and now it work, but it is no the right way.

enzolutions:
@mondrake could you help us with this, to see if we could include in next week release 1.9.1? Thanks in advance

mondrake:
@enzolutions @harold20 sure! What is exactly the problem here? Sorry but I can not understand.

mondrake:
@enzolutions @harold20 I have rebased and pushed my branch, now all checks on TravisCI are passing. What else you need?

mondrake:
@enzolutions @harold20 now, after #4136 and #4137, you can see in the TravisCI logs the effect of this PR on the `site:status` command

PR #4080, opened 2019-06-06
[site:install] #4079 Exit code issue fix

### LINKED ISSUE #4079: [site:install] Exitcode 0 even when some error occurs
### Problem/Motivation
The command should return a non-zero value when there is a problem like a php extension missing, or the site is already installed.

**Details to include:**
If we put the command on a CI pipeline, it is desirable to get a proper error code to avoid false positive situations.

### How to reproduce
* Run the `site-install` command on an already installed drupal site;
* run `echo $?` to see the error code;
* The proper error code should at least be 1, not 0;

**Details to include:**
- Console version: `1.8.0`.
- Console Launcher version: `1.8.0`.

### Solution
My suggestion is that we get the return value of the `runInstaller` in this line:  
https://github.com/hechoendrupal/drupal-console/blob/6af1543f914bfdec29497441783704eb0ef7483a/src/Command/Site/InstallCommand.php#L466  
And return it right away if it's not zero.

zoiosilva:
Fixes #4079

### DIFF
--- a/src/Command/Site/InstallCommand.php
+++ b/src/Command/Site/InstallCommand.php
@@ -463,7 +463,9 @@ class InstallCommand extends ContainerAwareCommand
         try {
             $drupalFinder = new DrupalFinder();
             $drupalFinder->locateRoot(getcwd());
-            $this->runInstaller($database_install, $uri);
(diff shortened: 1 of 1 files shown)


enzolutions:
Thanks @zoiosilva

PR #4106, opened 2019-06-28
Remove unused variable

gnikolovski:
- Variable $entity is not used, so PHP CS is throwing a warning about an unused variable.


enzolutions:
Thank you @gnikolovski

PR #4115, opened 2019-07-29
[debug:module] fix "modules" argument type mismatch

jensschulze:


jensschulze:
This is not even close to completed.
Will reopen when I have addressed the subsequent errors e.g. there is no GuzzleHttp\Client::getUrlAsJson() method

PR #4132, opened 2019-08-20
Fix TravisCI failure

mondrake:
PHP 5.5.9 runs on Trusty, add PHP 7.4, drop HHVM


mondrake:
Will submit a more comprehensive PR.

mondrake:
See #4136

Issue #4133, opened 2019-08-20
[user:create] shows error when --roles is specified.


ndouglas:
### Problem/Motivation
We see an error like this:
```
[ERROR] <em class="placeholder">Warning</em>: implode(): Invalid arguments passed in <em                               
         class="placeholder">Drupal\Console\Command\User\CreateCommand-&gt;createUser()</em> (line <em                  
         class="placeholder">249</em> of <em                                                                            
         class="placeholder">/var/www/drupalvm/vendor/drupal/console/src/Command/User/CreateCommand.php</em>). <pre     
         class="backtrace">Drupal\Console\Command\User\CreateCommand-&gt;createUser(&#039;username&#039;,             
         &#039;password&#039;, &#039;role&#039;, NULL, &#039;1&#039;) (Line: 119)                               
         Drupal\Console\Command\User\CreateCommand-&gt;execute(Object, Object) (Line: 255)                              
         Symfony\Component\Console\Command\Command-&gt;run(Object, Object) (Line: 987)                                  
         Symfony\Component\Console\Application-&gt;doRunCommand(Object, Object, Object) (Line: 255)                     
         Symfony\Component\Console\Application-&gt;doRun(Object, Object) (Line: 188)                                    
         Drupal\Console\Core\Application-&gt;doRun(Object, Object) (Line: 64)                                           
         Drupal\Console\Application-&gt;doRun(Object, Object) (Line: 148)                                               
         Symfony\Component\Console\Application-&gt;run() (Line: 89)                                                     
         require(&#039;/var/www/drupalvm/vendor/drupal/console/bin/drupal.php&#039;) (Line: 4)                          
         </pre>     
```

### How to reproduce
Attempt to create a user with the following syntax:

`drupal user:create <username> <password> --roles="<role>" --status=1 --no-interaction`

**Details to include:**
- Drupal version: 8.7.6
- Console version:  v1.9.0

### Solution
It looks like the `roles` option isn't transformed into an array at any point if it is provided as an argument.  However, in `createUser`, after the user is created, the code assumes that the `$roles` variable is an array and tries to `implode` it, which causes an error to be thrown while populating `$result['success']`, which then results in a nasty backtrace.

I'll submit a pull request here in a bit.

PR #4134, opened 2019-08-20
[user:create] Ensure $roles is an array when creating users.

ndouglas:
This prevents an error from being thrown when constructing the success message.  See #4133.


enzolutions:
Thank you @ndouglas

PR #4143, opened 2019-09-01
Add --default option to database:add

Kwadz:
This option allows to set the database as the default one instead of
adding a new database in addition to the default one.

See: https://github.com/hechoendrupal/drupal-console/issues/4142


enzolutions:
Thanks @Kwadz

PR #4153, opened 2019-09-17
Fix path config:export on root Drupal

gilbertomangones:
Fix path generating config:export. 

Generating the directory in root Drupal when is applied the next command.

```drupal config:export``` path relative
or 
```drupal config:export --directory=config2.test```
or 
```drupal config:export --directory="mod/config2.test"```

testin2


enzolutions:
Thanks @gilbertomangones for your contribution

PR #4155, opened 2019-09-17
Fix generate:entity:content installing module

gilbertomangones:
Hi,  ```generate:entity:content``` generate an error to the install.

Thanks


enzolutions:
Thanks @gilbertomangones

PR #4158, opened 2019-09-24
Add option install theme with release version

gilbertomangones:
This added feature works similarly to how "module: install" works, 
To install a theme using the release version when it is not found.

Thanks.


enzolutions:
Thanks @gilbertomangones

PR #4165, opened 2019-10-02
Fix failing TravisCI builds

mondrake:
Recent merges have caused TravisCI builds to fail.

Two methods with the same name were present in the Validator class, and a service definition was not updated. This PR fixes that.

Please, ensure that when you merge PRs, TravisCI is passing on the PR. Otherwise, there's no purpose to have that check.


enzolutions:
Thanks @mondrake

PR #4170, opened 2019-10-06
Removes constructor override on Controller

GueGuerreiro:
Addresses https://github.com/hechoendrupal/drupal-console/issues/4169 for the generate:controller command.

This would require https://github.com/hechoendrupal/drupal-console-core/pull/369 to be merged first.


GueGuerreiro:
I decided to create https://github.com/hechoendrupal/drupal-console/pull/4174 to bundle them all together, since I think that makes the most sense, as it avoids having to create multiple branches for each commit.

Closing this one.

PR #4175, opened 2019-10-09
[generate:plugin:block] Fixed block template using incorrect item and sequence order in 'for' loop.

AlexSkrypnyk:
https://twig.symfony.com/doc/2.x/tags/for.html


enzolutions:
Thanks @AlexSkrypnyk for your contribution, will be included in next release 1.9.4 in couple of weeks.

PR #4177, opened 2019-10-09
Use injected messenger service if possible.

### LINKED ISSUE #4176: [generate:form] When messenger service injected use it instead of \Drupal::messenger()
### Problem/Motivation

When using the following command to generate a form with the messenger service injected the generated form's submit function calls `\Drupal::messenger()` where it could call `$this->messenger` instead to comply Drupal coding standards.

### How to reproduce

```bash
drupal generate:form  \
--module="hello_world" \
--class="HelloWorldForm" \
--form-id="hello_world_form" \
--services='messenger' \
--path="/hello-world-form" \
--no-interaction
```

### Solution

These are the relevant lines in the corresponding template used to generate the form class:

https://github.com/hechoendrupal/drupal-console/blob/f26fd9b5fdb389719d77ff30849af714762e7d2b/templates/module/src/Form/form.php.twig#L114-L119

They could be changed to:

```
  public function submitForm(array &$form, FormStateInterface $form_state) {
    // Display result.
    foreach ($form_state->getValues() as $key => $value) {
{% if 'messenger' in services|keys %}
      $this->messenger->addMessage($key . ': ' . ($key === 'text_format'?$value['value']:$value));
{% else %}
      \Drupal::messenger()->addMessage($key . ': ' . ($key === 'text_format'?$value['value']:$value));
{% endif %}
    }
  }
```

leymannx:
Closes #4176

### DIFF
--- a/templates/module/src/Form/form.php.twig
+++ b/templates/module/src/Form/form.php.twig
@@ -114,7 +114,11 @@ class {{ class_name }} extends FormBase {% endblock %}
   public function submitForm(array &$form, FormStateInterface $form_state) {
     // Display result.
     foreach ($form_state->getValues() as $key => $value) {
+{% if 'messenger' in services|keys %}
+      $this->messenger->addMessage($key . ': ' . ($key === 'text_format'?$value['value']:$value));
+{% else %}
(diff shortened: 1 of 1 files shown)


enzolutions:
Thanks @leymannx

PR #4178, opened 2019-10-10
[module:install] Remove moduleRequirement()

mondrake:
Module/InstallCommand calls a pre-flight check of the module requirements. That check is embedded into a method in the class; however, that method calls old APIs, no longer runs after Drupal 8.7.7, and is largely duplicating the checks that Drupal's module installer service is performing upon module installation.

This PR just removes that method and its invocation.


enzolutions:
Thanks @mondrake

PR #4181, opened 2019-10-14
[generate:entity:content] offers per bundle permissions for no-bundle entity

nequeteme:
validate if the entity has bundles it's on true and create permissions https://github.com/hechoendrupal/drupal-console/issues/4162


enzolutions:
Thanks @nequeteme

PR #4183, opened 2019-10-14
[user:login:url] validate https on uri

nequeteme:
Validate if the uri have https and return the link with https
https://github.com/hechoendrupal/drupal-console/issues/3935


enzolutions:
Thanks @nequeteme

PR #4187, opened 2019-10-25
[generate:theme] fix libraries.yml syntax

lalop:
There is an indexation error in the template, the generated file is invalid


enzolutions:
Thasks @lalop

Issue #3930, opened 2018-09-16
[site:install] Distinguish between an empty db prefix and no db prefix


jensschulze:
### Distinguish between an empty db prefix and no db prefix

[site:install] Atm you have to use `--no-interaction` if you want to pass an empty db prefix. If you type `--db-prefix=""` the result is the same as if you had just simply omitted this option.

### Problem/Motivation
If the value of an option is explicitly set it should not be asked for later. You can suppress this inconsistent behaviour with `--no-interaction` but one should not need 
I am running into this issue whenever I'm trying to set up a "throw away" installation.

### How to reproduce
Last checked on Drupal 8.6.1 with drupal/console 1.8.0

```bash
composer create-project drupal-composer/drupal-project:8.x-dev somedirectory --stability dev --no-interaction

cd somedirectory

vendor/bin/drupal si standard \
    --langcode="en" \
    --db-type="sqlite" \
    --db-file="sites/default/files/drupaldb.sqlite " \
    --db-prefix="" \
    --account-name=admin \
    --account-pass=admin \
    --account-mail=admin@example.com \
    --site-name="This is just a test" \
    --site-mail="admin@example.com"
```

Even though the `--db-prefix` is set, the command will ask for the prefix (the default value is an empty string).

### Solution
If no `--db-prefix` is given, the option's default value is _null_. So in Site\InstallCommand::interact() we can easily distinguish between "no `--db-prefix` given" (ask) and "empty `--db-prefix` given" (no action).

LOBsTerr:
@jensschulze Thanks I have merged it

PR #4189, opened 2019-11-05
[site:install] Regression: command does not respect empty db-prefix

jensschulze:
### Problem/Motivation
The issue that was addressed in commit e62a1fe #3930 #3931 is back.

If you submit an empty string as DB prefix (`--db-prefix=''`) it is handled exactly as if no db-prefix was given: the command asks for a prefix regardless of the input. This is due to a regression first introduced by f1dc3aa #3917. This regression was partly reverted by 1247f20 #4049, but in both cases **PHP’s implicit type casting** causes problems.

You have to check the input value explicitly against `null`, otherwise a blank string ("no prefix") will be handled like "no value given".

```php
if (!$something) {
    echo '$something is either null or 0 or false or an empty array or an empty string. Are you sure this is the desired behavior?';
}
```

### How to reproduce
- Drupal version: tested on Drupal 8.6+
- Console version: Regression was introduced in 1.9.1

```shell-script
drupal -y si test \
    --langcode="en" \
    --db-type="mysql" \
    --db-host="testdbhost" \
    --db-port="3306" \
    --db-name="testdb" \
    --db-user="testuser" \
    --db-pass="testpass" \
    --db-prefix="" \
    --account-name=admin \
    --account-pass=admin \
    --account-mail=admin@example.com \
    --site-name="Test (MySQL)" \
    --site-mail="admin@example.com"
```

Result:

```plaintext
 Database Prefix []:
 >
```

### Solution
Avoid implicit type casting and check `$input->getOption('db-prefix')` against `null` explicitly.

### DIFF
--- a/src/Command/Site/InstallCommand.php
+++ b/src/Command/Site/InstallCommand.php
@@ -323,7 +323,7 @@ class InstallCommand extends ContainerAwareCommand
             }
 
             // --db-prefix option
-            if (!$input->getOption('db-prefix')) {
+            if ($input->getOption('db-prefix') === null) {
                 $input->setOption('db-prefix', $this->dbPrefixQuestion());
             }
         }


enzolutions:
Thanks @jensschulze

PR #4191, opened 2019-11-08
[console] Add an extra path to find the autoload

harold20:


enzolutions:
@novia713 this PR should fix your issues, I will do a new release tomorrow, you need to test everything before your session.

PR #4174, opened 2019-10-08
Removes constructor override on service object generators.

GueGuerreiro:
This addresses https://github.com/hechoendrupal/drupal-console/issues/4169. 

It requires https://github.com/hechoendrupal/drupal-console-core/pull/369 to be merged first, for a few of the generators that use the function Twig function `serviceClassInjection`.

PR #4192, opened 2019-11-09
Bugfix for #4174.

GueGuerreiro:
There was a bug on some of the templates on the previous PR I made.
When the class itself is implementing ContainerFactoryPluginInterface, there's no parent create method we can use. We should fallback to the class __construct() method instad.
This PR does just that.


enzolutions:
Thanks @GueGuerreiro , just in time for the release 1.9.4

PR #4199, opened 2019-11-20
fix(generator): block description index

sylvainmetayer:
`block_description` is undefined, the one to use is `description`


LOBsTerr:

PR #4208, opened 2019-11-29
fix: missing property

sylvainmetayer:
When calling `drupal generate:event:subscriber` : 

```
In EventsTrait.php line 59:
Undefined property: Drupal\Console\Command\Generate\EventSubscriberCommand::$events
```

The variable could either be in the trait or in the class using it. From what I've seen, it is only used in the EventSubscriberCommand class.


LOBsTerr:

PR #4215, opened 2019-12-17
Update links.task-entity-content.yml.twig

Holo96:
Field UI link tasks like 'manage fields' are using this link task as base route, so without this you can never switch back to bundle edit tab from other tabs
entity-local-tasks


LOBsTerr:

PR #4216, opened 2019-12-19
[generate:entity:content] per-bundle-permissions

hudri:
[generate:entity:content] wrongly creates one permission per entity (database row), instead of one permission per entity type.


severin19:
Applying the changes in this commit resolved the issue for me.  I would be good to get this merged for future use.

LOBsTerr:

Issue #4195, opened 2019-11-14
[config:export] Path is broken on windows


sylvainar:
### Problem/Motivation

I can't export my config because of broken path. That's a big issue since `config:export` / `config:import` are two of my most used commands.

```
In Filesystem.php line 104:

  Failed to create "D:\Profiles\myname\PhpstormProjects\myproject/D:\Profiles\myname\PhpstormProjects\myproject/../sync": mkdir(): No such file or directory.
```

### How to reproduce

Install module, then export config.

- OS: Windows 10
- Drupal version: 8.7.10
- Console version: 1.9.4
- Console Launcher version: 1.9.4
- Steps to reproduce: `drupal config:export`

![image](https://user-images.githubusercontent.com/9823286/68857104-654ad500-06e2-11ea-9b27-3dbd1e8a3e67.png)

![image](https://user-images.githubusercontent.com/9823286/68857193-95927380-06e2-11ea-8e96-68fcdc27593a.png)

### Solution

Dunno but it seems related to this issue: https://github.com/hechoendrupal/drupal-console/pull/4154

Quick fix for now has been to rollback to `1.9.2`.

timotheemoulin:
I've looked into it (had the same issue) and fixed it in a small patch.

LOBsTerr:
Thank you for your contribution

PR #4219, opened 2019-12-27
fix sync directory path

timotheemoulin:
This pull request is fixing the issue reported here

#4195 [config:export] Path is broken on windows #4195
https://www.drupal.org/project/drupal/issues/3103359


B-Prod:
Actually it seems that the change makes the config to be saved on the parent folder of the defined path.
For example with : `../config/default/sync`, the config is not stored to the parent folder of the Drupal root, but to the level 2 parent.

PR #4237, opened 2020-03-17
[config:import:single] Prevent writing to DB before accessing ConfigImporter.

### LINKED ISSUE #4236: [config:import:single] Writing to DB before using the ConfigImporter.
### Problem/Motivation

We noticed this issue when trying to install modules using the config:import:single command on a core.extension.yml file. While trying to install a module with this command the install hook is skipped which will result in missing DB tables.

This issue is caused by line 106 of the src/Command/Config/ImportSingleCommand.php file `$source_storage->write($name, $value);` 

This write is hitting the write function of the underlaying DatabaseStorage object, which means it is already written to the DB before the ConfigImporter class is able to make it's comparison. When the ConfigImporter is initialised the Source and Target storages are exactly the same (Since they are both reading from the same DB).

I used xdebug while running this command and found that both the `totalExtensionsToProcess` and  `totalConfigurationToProcess` properties of the ConfigImporter are always 0, which means the ConfigImporter is never writing to the DB.

To verify my theory I removed the entire `$this->importConfig()` block which gave me the exact same end result.

### How to reproduce

Import a core.extension.yml file which includes a new module with it's own DB table. Afterwards the DB table will be missing.

**Details to include:**

System Info

Property | Value
--- | ---
Drupal | 8.8.1
Access to update.php | Protected
Configuration files | protected
Cron maintenance tasks | Last run 4 minutes 2 seconds ago
Database system | MySQL, MariaDB, Percona Server, or equivalent
Database system version | 5.6.43
Database updates | Up to date
Drupal core update status | No update data available
Entity/field definitions | Up to date
File system | Writable (public download method)
JSON:API allowed operations | Read-only
JSON:API revision support | Limited
Node Access Permissions | Disabled
PHP | 7.3.11 (more information)
PHP extensions | Enabled
PHP OPcode caching | Enabled
Random number generation | Successful
Search index progress | 100% (0 remaining)
Serialized data types | Not enabled
Trusted Host Settings | Not enabled
Unicode library | PHP Mbstring Extension
Update notifications | Enabled
Upload progress | Not enabled
Drupal Console | 1.9.4

### Solution
The solution is quite simple, we can replace line 105 and 106:

```php
$source_storage->delete($name);
$source_storage->write($name, $value);
```

with: 

```php
$source_storage->replaceData($name, $value);
```

Looking at the file history it seems that the initial implementation was using this function, so not sure why it was changed. The commit which changes this has as far as I can see no reason to change this line.

I am currently creating a Patch file for our own project so will create a PR later today. (Assuming this is a bug and not intended behaviour).

melvinversluijs:
Closes #4236

### DIFF
--- a/src/Command/Config/ImportSingleCommand.php
+++ b/src/Command/Config/ImportSingleCommand.php
@@ -102,8 +102,7 @@ class ImportSingleCommand extends Command
                     $name = Path::getFilenameWithoutExtension($configFile);
                     $ymlFile = new Parser();
                     $value = $ymlFile->parse(file_get_contents($configFile));
-                    $source_storage->delete($name);
-                    $source_storage->write($name, $value);
+                    $source_storage->replaceData($name, $value);
                     $names[] = $name;
                     continue;
                 }


LOBsTerr:

PR #4239, opened 2020-03-26
Fix fatal when scaffolding a FieldWidget.

Artusamak:
If you generate a new field widget through the scaffolding command `drupal gpfw`, you can't use your widget right away due to a fatal error caused by a missing comma and annotations inclusion.

The attached PR fixes it.


LOBsTerr:

PR #4246, opened 2020-05-19
[config:export] fix path construction (#4245)

### LINKED ISSUE #4195: [config:export] Path is broken on windows

### LINKED ISSUE #4198: [config:export] broken in 1.9.4
[config:export] broken in 1.9.4

### Problem
After upgrading to console 1.9.4, `config:export` incorrectly calculates the directory path for the export files, creates those incorrect directories, and doesn't export any files. No error is thrown.

May be related to https://github.com/hechoendrupal/drupal-console/issues/4195.

I am using a web docroot on Ubuntu 18.04 local development environment (Lando/Pantheon recipe), and

* Drupal 8.7.10
* PHP 7.2
* Console 1.9.4

**What I expect to see**: execute `drupal ce`, accept the default directory (`/app/config`). All config files are exported to `/config`. (Note with a web docroot this is parallel to the `web` directory containing the drupal site). This worked as expected until upgrading to 1.9.4.

**What happens instead**:
image

Which creates an `/app/app/config` directory with no files in it:
image

Replacing the default with `/config` generates an `/app//config` directory structure with no files in it.

Replacing the default with `/` destroys the site. Do not recommend. 

### How to reproduce
* Upgrade drupal/console to 1.9.4
* Attempt a config:export

### Workaround
Revert to 1.9.3, which works as expected

### LINKED ISSUE #4245: [config:export] directory incorrectly prefixed with composer root
### Problem/Motivation

In #4154 / 37aeb5d12, ExportCommand was modified to prefix the config sync directory with the composer root. Three bugs:

1. A relative path is relative to *Drupal* root, which is not the same as the composer root - many sites created via composer put Drupal in a `./web/` subdirectory.
2. The composer root is actually prefixed twice when making the directory.
3. The path might already be absolute, in which case it should not be prefixed with anything at all.

### How to reproduce
Include steps related how to reproduce.

- Create a Drupal site using the `drupal-composer/drupal-project` template (https://github.com/drupal-composer/drupal-project)
- Require `drupal/console` ^1.9.4
- Set `$settings['config_sync_directory'] = '../config/sync';` in `settings.php`.
- Run `drupal config:export`

Expected:
- Config is exported to `config/sync` from composer root.

Actual:
several issues (getting a different one on each environment where I reproduced it):

1. Tries to create `{COMPOSER_ROOT}/{COMPOSER_ROOT}/../config/sync`
2. Tries to export to `{COMPOSER_ROOT}/../config/sync`

### Solution
A brief description of the proposed fix.

- Use Webmozart\PathUtil\Path to check if the directory is absolute
- If not, use `drupalFinder->getDrupalRoot()` and prefix the directory with that
- Remove the extra prefixing when creating the directory.

(PR to follow)

cburschka:
- Use Drupal root, not Composer root.
- Avoid adding the root to an already absolute path.
- Avoid adding the root twice when creating the directory

Fixes #4245, fixes #4195, fixes #4198, fixes #4219

### DIFF
--- a/src/Command/Config/ExportCommand.php
+++ b/src/Command/Config/ExportCommand.php
@@ -17,6 +17,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Drupal\Console\Core\Command\Command;
 use Symfony\Component\Filesystem\Filesystem;
 use Drupal\Core\Config\ConfigManager;
+use Webmozart\PathUtil\Path;
 
 class ExportCommand extends Command
 {
@@ -94,20 +95,22 @@ class ExportCommand extends Command
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $drupal_root = $this->drupalFinder->getComposerRoot();
-        $directory = $drupal_root.'/'.$input->getOption('directory');
+        $directory = $input->getOption('directory');
         $tar = $input->getOption('tar');
         $removeUuid = $input->getOption('remove-uuid');
         $removeHash = $input->getOption('remove-config-hash');
-        $drupal_root = $this->drupalFinder->getComposerRoot();
 
         if (!$directory) {
             $directory = config_get_config_directory(CONFIG_SYNC_DIRECTORY);
         }
+        if (!Path::isAbsolute($directory)) {
+            $drupal_root = $this->drupalFinder->getDrupalRoot();
+            $directory = $drupal_root . "/" . $directory;
+        }
 
         $fileSystem = new Filesystem();
         try {
-            $fileSystem->mkdir($drupal_root."/".$directory);
+            $fileSystem->mkdir($directory);
         } catch (IOExceptionInterface $e) {
             $this->getIo()->error(
                 sprintf(


cburschka:
Note: The third of the fixes in this PR is also contained in #4219, and should be credited to @timotheemoulin.

timotheemoulin:
I'm glad someone else took the lead on this issue

PR #4272, opened 2020-10-12
Add package path option.

verbruggenalex:
Patch proposal to add package-path option to allow the generation of modules outside of the Drupal webroot: https://github.com/hechoendrupal/drupal-console/issues/4118#issuecomment-707162051


felubra:
@verbruggenalex thanks for the patch, it works. Could somebody review this? Thanks.

LOBsTerr:

PR #4309, opened 2021-07-07
fix optional argument before mandatory argument php8 warning

### LINKED ISSUE #4310: [console] PHP warning in DrupalServiceModifier when commands called on PHP8
### Problem/Motivation

PHP8 outputs a notice for methods that define optional arguments before required arguments (https://php.watch/versions/8.0/deprecate-required-param-after-optional).

### How to reproduce
- Drupal version: 9.1.10
- Console version: 1.9.7

Run any console command on a php8 installation. E.g.
1. download and run docker image of Drupal 9.1
2. install drupal-console using composer
3. attempt to call a command.

### Solution
Remove the default value assignment for the first parameter. Pull request opened.

jenswegar:
PHP8 outputs a notice for methods that define optional arguments before required arguments (https://php.watch/versions/8.0/deprecate-required-param-after-optional). Since the DrupalServiceModifier constructor in effect cannot be called without also providing the first argument (even if it's null), it should be safe to just remove the default value assignment. 

Fixes #4310

### DIFF
--- a/src/Bootstrap/DrupalServiceModifier.php
+++ b/src/Bootstrap/DrupalServiceModifier.php
@@ -39,7 +39,7 @@ class DrupalServiceModifier implements ServiceModifierInterface
      * @param ConfigurationInterface $configuration
      */
     public function __construct(
-        $root = null,
+        $root,
         $serviceTag,
         $generatorTag,
         $configuration


Luigisa (approved):

rjbain:
Bump.

avioli:
bump bump

LOBsTerr:
Was fixed in this PR - https://github.com/hechoendrupal/drupal-console/pull/4327

THanks for your contribution

PR #4334, opened 2022-09-07
fix: Use service interfaces, not specific implementations.

### LINKED ISSUE #4333: fix: Use ModuleHandlerInterface instead of ModuleHandler
Currently the implementation of `Update/ExecuteComamnd` relies on `ModuleHandler`, however some modules might want to decorate/replace this service so the class name won't match.

https://github.com/hechoendrupal/drupal-console/blob/d292c940c07d164e32bbe9525e909311ca65e8cb/src/Command/Update/ExecuteCommand.php#L80

esolitos:
This is just a simple fix which replaces the references to `\Drupal\Core\Extension\ModuleHandler` with it's respective interface `\Drupal\Core\Extension\ModuleHandlerInterface`.

Fixes #4333


LOBsTerr:
Thanks for contribution

PR #4348, opened 2023-09-07
Issue #3337542: Drupal 10 compatibility fixes

erickmajor:
The current module version isn't able to use on Drupal 10 implementations.

Some Drupal community members opened the https://www.drupal.org/project/console/issues/3337542 issue, reporting this compatibility issue and brought some fixes. These code modifications were brought by these members.

Please, update this module to cover the new Drupal's version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,16 +43,16 @@
     "require": {
         "php": ">=7.0.8",
         "alchemy/zippy": "~0.4",
-        "composer/installers": "~1.0",
+        "composer/installers": "^1.2",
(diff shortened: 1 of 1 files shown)


MdNadimHossain:
how can we escalate this to be merged. I am using forked version get around the issue for my D10 upgrades.

ludal74:
The fix is not relevant for official Drupal core composer support through package "drupal/recommended-project" which requires composer/installers ^2.0 for core version 10 and above.

Please consider upgrading the dependency to that version :

 "composer/installers": "^2.0",

